PR #117, opened 2023-06-12
Consolidate Windows and Linux CI config files

### LINKED ISSUE #115: Consolidate Windows and Linux GitHub Actions config files
Currently we have two files, `windows.yml` and `linux.yml`, for the two different build actions we're running on the CI, with some duplicated content. This is fine, but it would be cleaner to have everything in a single config file (with the OS-specific parts guarded with `if`s as appropriate) so that things can just be updated in one place.

Note that this will become even more of a problem if/when we add additional targets (see e.g. #84, #81, #76), so we should probably do this first.

alexdewar:
This will make it easier for us to add additional checks and targets to the CI config in future.

I've been quite conservative, in that I've tried to leave the configuration basically the same except for merging `linux.yml` and `windows.yml` (though I did remove some unnecessary steps).

Fixes #115.


alexdewar:
PS -- I'm think that after this PR is merged we should issue a new point release just to check I haven't broken any of those jobs.

alexdewar:
Hi @dalonsoa. I thought it might be helpful to have your input on this too, if that's ok. Also, if @cc-a could have a look over it that'd be great (I couldn't add you as a reviewer presumably because you're not part of the org).

I've broadly kept the same structure that they had before for the CI config files, but I've just consolidated the separate Windows and Linux files into a single one (using `matrix.os`). As before though, all of the tasks -- i.e. building and testing -- are run as part of a single mega-job, whereas it'd probably be nicer to have them appear as separate jobs (to make it more obvious what's broken when it fails etc.). The problem is that the "test" step needs the files produced in the "build" step in order to work. It seems like you *can* copy files between jobs by caching them, but this seems rather hacky. I was just wondering if either of you have any ideas about how we could do this more cleanly?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,93 @@
...
+        cd vcpkg
+        ./bootstrap-vcpkg.sh -disableMetrics
+
+    - name: Build HealthGPS (debug)
+      run: |
+        cmake --preset=${{ matrix.platform }}-debug -DBUILD_BENCHMARKS=ON
+        cmake --build --preset=debug-build-${{ matrix.platform }} --target=install 
+
+    - name: Build HealthGPS (release)
+      run: |
```
dalonsoa, line 58: Why are we building both the debug and release versions every single time?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,93 @@
...
+        mkdir ${{github.workspace}}/artifact
+        7z a -tzip ${{github.workspace}}/artifact/health_gps_${{ matrix.platform }}.zip *.dll *.Console*
+
+    - name: Upload release artifacts
+      if: startsWith(github.ref, 'refs/tags/')
+      uses: softprops/action-gh-release@v1
+      with:
+          files: artifact/health_gps_${{ matrix.platform }}.zip
+
+    - name: Doxygen release API deploy
```
dalonsoa, line 85: If creating the documentation does not require any of the previous steps - except fetching the repo - i would separate it into a different job, so it can run in parallel with the testing ones.

README.md
```diff
@@ -1,5 +1,4 @@
-![windows](https://github.com/imperialCHEPI/healthgps/actions/workflows/windows.yml/badge.svg)
-![Linux](https://github.com/imperialCHEPI/healthgps/actions/workflows/linux.yml/badge.svg)
+[![CI](https://github.com/imperialCHEPI/healthgps/actions/workflows/ci.yml/badge.svg)](https://github.com/imperialCHEPI/healthgps/actions/workflows/ci.yml)
```
dalonsoa, line 1: This might be the reason to keep it separated into two workflows, so it is explicit in which platform each thing runs. We know it is a purely cosmetic thing and that the tests are running in both, anyway, but I guess it might be relevant to consider it?

dalonsoa (approved):
Ive just left a couple of really minor, not sure if relevant, suggestions. Otherwise, I thinks looks good and that is a sensible move towards having shorter, more concise code.

.github/workflows/ci.yml
alexdewar: That's what they were doing before, so I kept it in.

In *theory* you could have build failures that only happen when you compile in debug mode rather than release or vice versa, but it's not particularly likely and it doubles the amount of time the CI spends building things. In my last job, we only build-tested in release mode and that worked fine.

I'd be inclined to open an issue for this and change it in another PR, so that this one is a purely non-functional change, but if we're happy that this is the correct course of action, we could just do it now.

@jamesturner246?

.github/workflows/ci.yml
alexdewar: The thing is, it really shouldn't, except that they've integrated building the docs with CMake, so it does kinda have to come after invoking CMake. Really though you should just be able to run `doxygen` manually though to just get the docs (it's fine to *also* have CMake optionally build the docs if the user wants, but it's annoying to have it as a hard dependency).

They're definitely not alone in doing this, but it probably makes sense to break that dependency. Again, maybe I should open an issue and make a new PR for it?

README.md
alexdewar: Yeah, I was disappointed to find that after all this work there was no way to get separate badges Linux and Windows here :disappointed:. (It would be even nicer if we could get separate badges for the building and testing steps too.)

My feeling is that losing the separate badges is probably just the price we have to pay for this PR. It may be that in future we can figure out how to separate out the various CI steps and get the badges back. We should enforce tests passing before PRs can be merged anyway, so hopefully it should only ever be green in any case!

@jamesturner246 maybe we should have a chat about what settings we want for the repo in terms of conditions for PRs being merged etc.?

.github/workflows/ci.yml
jamesturner246: The build process is currently not so long anyway, but happy to leave it for now.

README.md
jamesturner246: Sure. At least testsuite gating, and maybe some valgrind and/or code linting. Happy to discuss ofc.

jamesturner246 (approved):
The Linux commands make sense, but can't comment on the Windows config side. Looks sensible enough -- go ahead.

.github/workflows/ci.yml
alexdewar: Cool. I've opened an issue for future reference: #122

PR #125, opened 2023-06-19
Add `pre-commit` config and check hooks in CI

### LINKED ISSUE #124: Add some `pre-commit` hooks and enforce with CI
This is partly a QoL improvement for developers -- in that `pre-commit` hooks can catch simple mistakes before they are committed -- but it's also an extra QA check we can run as part of the CI.

alexdewar:
I've currently just got a few hooks in there. I was trying to be relatively conservative with my choices, but we can add more later if we want. I've incorporated the `pre-commit` checks into the CI too (so we get two passes at catching errors) and I've added another GitHub Action for auto-updating the hooks.

Essentially I've added the minimum amount of infrastructure that we'll need to add other hooks (such as one for `clang-format`). I stopped short of adding hooks which would reformat large numbers of files so as not to make the diff large and messy.

Closes #124.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,8 +93,24 @@ cmake --build --preset='debug-build-linux'
 ctest --preset='core-test-linux'
 ```
 
-All available options are defined using CMake *presets* in the `CMakePresets.json` file, which also declare *build presets* to be used via the CMake command line arguments.
+All available options are defined using CMake *presets* in the `CMakePresets.json` file,
+which also declare *build presets* to be used via the CMake command line arguments.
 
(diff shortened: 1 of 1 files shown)


jamesturner246 (approved):
Yes, this looks fine. Just an aside, but is there any real difference/necessity to run any of them on Windows? Thanks.

alexdewar:
There's no need to run the `pre-commit` hooks on Windows, as they should behave exactly the same as on Linux. For building and testing, things might differ which is why we also run those steps on Windows.

PR #129, opened 2023-06-20
New Energy Balance Model input layer: computing nutrients from food groups

### LINKED ISSUE #112: New Layer 1 (input) for EBM
# Feature description

We want to generalise the input to Layer 1 (fat, carbs, Na, etc) to handle both food groups and base nutrients as input data using a similar generic mechanism. i.e. we want to be able to input something like:

|food group | Alcohol |   Carbs  |  Fats   | Protein |   Sodium |
|-|-|-|-|-|-|
|pizza  |    0  |       20 |        30  |   10   |        10|
|salad   |   0     |    2     |      1     |   2        |    20|
|chicken |   0    |     2    |       20  |    30    |       10|
|carbs  |   100    |    0   |      0   |   0      |        0|
|protein    |   0       |  0       |     0    |  100   |       0|

Note the "food groups" carbs and protein respectively contain nothing but 100% carbs and 100% protein, whereas other food groups contain mixtures of nutrients.

# Implementation

This could involve the implementation of an additional 'layer 0', which will take as input each food group risk factor, and output the accumulated nutrients of all food group risk factors. These would be input to the existing 'layer 1', which expects nutrients, rather than the starting food groups.

In the input `.Config.json` file: entries in `"modelling": { "risk_factors"` should instead be a food group (either a literal food group or a pure nutrient entry like fat), and parameters/coefficients for converting from food groups to nutrients must be provided in the dynamic configuration json files (`France.EBHLM.json` in the healthgps example):

```
...
"Equations": {
    "0-19":{
        "Female":[
...
            {
                "Name":"Pizza"
                "Coefficients":{
                     "Intercept":...
                     "SES":...
                     "Age":...
                },
                "ResidualStandardDeviation":...
            },
...
            {
                "Name":"Fat"
                "Coefficients":{
                     "Intercept":...
                     "Fat":...
                },
                "ResidualStandardDeviation":...
            },
...
```

etcetera.

Also, in initialisation parameters should be provided in the static model json file (`France.HLM.json` in the healthgps example) for *both* the food groups (layer 0) *and* the nutrient groups (layer 1), since both layers will need to be initialised.

Thoughts on this?

jamesturner246:
This change adds the functionality and configuration files for the new Energy Balance model, allowing it to compute nutrient sums from input food groups using coefficients provided in a new config file.

The user can still choose between this and the old model by specifying the model's config file in `modelling -> risk_faxctor_models -> dynamic` as before. The config file for the new dynamic model is `./example/France.NewEBM.json`.

When Health-GPS initialises, and constructs the dynamic model, the correct dynamic model class is selected based on `ModelName` keyword in the dynamic config  `./example/France.NewEBM.json` (the old `./example/France.EBHLM.json` file now also now has a `ModelName` entry to allow it to be correctly loaded).

The new config file currently allows users to specify a list (`Nutrients`) of nutrients to be computed from the input food groups, and also an array (`Foods`) of dicts for each food group, specifying their nutrient coefficients. These are loaded by the new EBM to compute total nutrient sums.

Currently the 'food groups' are themselves nutrients, so these coefficients are simply 1.0 for the nutrient represented by the food group, and 0.0 otherwise (see #112 for a description).

Fixes #112


src/HealthGPS/energy_balance_hierarchical_model.h
```diff
@@ -0,0 +1,43 @@
...
+#include "random_algorithm.h"
+
+namespace hgps {
+
+	/// @brief Implements the dynamic hierarchical linear model (energy balance) type
+	/// 
+	/// @details The dynamic model is used to advance the virtual population over time.
+	class EnergyBalanceHierarchicalModel final : public HierarchicalLinearModel {
+	public:
+		EnergyBalanceHierarchicalModel() = delete;
```
alexdewar, line 13: I don't *think* you need this.

src/HealthGPS/hierarchical_model_registry.h
```diff
@@ -35,4 +39,15 @@ namespace hgps {
...
+	/// @brief Gets the default production instance builders for energy balance models
+	/// @return The energy balance models builder functions registry
+	std::map<HierarchicalModelType, detail::EnergyBalanceModelBuilder> get_default_energy_balance_model_registry() {
+		auto registry = std::map<HierarchicalModelType, detail::EnergyBalanceModelBuilder>{
+			{HierarchicalModelType::Dynamic, [](EnergyBalanceModelDefinition& definition) {
+				return std::make_unique<EnergyBalanceModel>(definition); }},
+		};
+
+		return registry;
+	}
```
alexdewar, line 52: All this ``*ModelBuilder`` stuff is something that we should be able to factor out at some point... This is a particularly cumbersome design :disappointed:

src/HealthGPS/repository.h
```diff
@@ -92,6 +98,14 @@ namespace hgps {
 			const HierarchicalModelType& model_type,
 			LiteHierarchicalModelDefinition&& definition);
 
+		/// @brief Register a user provided energy balance model definition
+		/// @param model_type The energy balance model type
+		/// @param definition The energy balance model definition instance
+		/// @return true, if the operation succeed; otherwise, false.
```
alexdewar, line 104:
```suggestion
		/// @return true, if the operation succeeds; otherwise, false.
```

src/HealthGPS/riskfactor.cpp
```diff
@@ -70,28 +70,30 @@ namespace hgps {
...
-		else {
-			models.emplace(HierarchicalModelType::Dynamic, lite_registry.at(HierarchicalModelType::Dynamic)(
-				repository.get_lite_linear_model_definition(HierarchicalModelType::Dynamic)));
+		catch (const std::out_of_range &oor) {
+			auto &dynamic_definition = repository.get_energy_balance_model_definition(HierarchicalModelType::Dynamic);
+			auto dynamic_builder = [](EnergyBalanceModelDefinition& dynamic_definition) {
+				return std::make_unique<EnergyBalanceModel>(dynamic_definition);
+			};
+			models.emplace(HierarchicalModelType::Dynamic, dynamic_builder(dynamic_definition));
 		}
```
alexdewar, line 97: I'd rather avoid this construction if possible, but I don't mind whether you do it in this PR or add it as an issue to be done some other time.

One option would be to have ``get_lite_linear_model_definition()`` return a ``std::optional<TheModelType &>>`` instead, with ``nullopt`` signalling that the model isn't available.

alexdewar (changes requested):
I've made some small suggestions, but on the whole it's looking good :smile:

src/HealthGPS.Console/model_parser.cpp
```diff
@@ -47,83 +47,70 @@ namespace host
...
-						.residuals_standard_deviation = at.residuals_standard_deviation,
-						.rsquared = at.rsquared
-						});
-				}
+		HierarchicalModelInfo model_info;
+		model_info.models = opt["models"].get<std::unordered_map<std::string, LinearModelInfo>>();
+		model_info.levels = opt["levels"].get<std::unordered_map<std::string, HierarchicalLevelInfo>>();
+
+		for (auto& model_item : model_info.models) {
+			auto& at = model_item.second;
```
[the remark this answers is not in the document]
jamesturner246: I'll fix these, and await the magic of `clang-tidy` for the rest!

src/HealthGPS/hierarchical_model_registry.h
jamesturner246: Agreed. in fact this whole nasty file can go with the new changes.

src/HealthGPS.Console/model_parser.cpp
```diff
@@ -132,108 +119,171 @@ namespace host
...
+			for (const std::string &nutrient_str : nutrient_strings) {
+				auto nutrient_ident = core::Identifier(nutrient_str);
+
+				try {
+					double val = food_nutrient_strings.at(nutrient_str);
+					nutrient_equations[food_ident][nutrient_ident] = val;
+				}
+				catch (const std::out_of_range &oor) {
+					continue;
+				}
```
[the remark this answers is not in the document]
jamesturner246: What have you done to me, Python -_-

src/HealthGPS.Console/model_parser.cpp
```diff
@@ -132,108 +119,171 @@ namespace host
...
 			}
 			else {
-				fmt::print(fg(fmt::color::red), "Unknown hierarchical model type: {}.\n", model.first);
-				continue;
+				fmt::print(fg(fmt::color::red),
+				"Model file: {} not found.\n", model_filename);
 			}
+
+			ifs.close();
```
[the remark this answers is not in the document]
jamesturner246: Ah. What's even the point of close? I guess if you're passing one around with unclear

src/HealthGPS/riskfactor.cpp
jamesturner246: Let's leave this for another PR. I suspect this will all change soon anyway.

src/HealthGPS/energy_balance_hierarchical_model.h
jamesturner246: Any reason not to delete default constructors in stateful classes in general?

src/HealthGPS/riskfactor.cpp
alexdewar: Cool. I'd open an issue for it so it doesn't get forgotten.

alexdewar: Cool. I'd open an issue for it so it doesn't get forgotten.

alexdewar:
I've just noticed there don't seem to be any restrictions on merging PRs atm. Usually we'd change the settings so that at least you can only do it if it has been approved by someone else and the tests are passing. Not a biggie, but maybe something to change at some point, to avoid accidental merges.

alexdewar:
PS -- feel free to click "request review" again whenever you're ready for it.

src/HealthGPS.Console/model_parser.cpp
```diff
@@ -47,83 +47,70 @@ namespace host
...
+				});
 		}
 
-		ifs.close();
 		return HierarchicalLinearModelDefinition{ std::move(models), std::move(levels) };
 	}
 
-	LiteHierarchicalModelDefinition load_dynamic_risk_model_info(std::string model_filename)
+	LiteHierarchicalModelDefinition load_dynamic_risk_model_definition(
+		const host::poco::json &opt)
```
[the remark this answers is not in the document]
jamesturner246: Kinda on the fence with the function name length debate. Agree that there's such thing as too long, but I like fn names that are somewhat self-documenting. There's probably a middle ground there..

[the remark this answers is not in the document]
jamesturner246: I wonder if it's even worth just abstracting the whole thing into a ModelLoader class for each model type.

alexdewar (approved):
LGTM! Nice work :-).

Merge at will...

jamesturner246:
Fantastic.  Thanks Alex! 😊

PR #138, opened 2023-06-23
Add `clang-format` `pre-commit` hook and reformat C++ files

### LINKED ISSUE #128: Use `clang-format` for whitespace formatting and add `pre-commit` hook
While the whitespace in the C++ files isn't badly formatted per se (though the style is a little idiosyncratic), there is currently no agreed standard or tooling in place for the repository. `clang-format` is the standard tool for this purpose and most IDEs and editors (including VS Code) will use it to format source code if a `.clang-format` file is present.

There is also a `pre-commit` hook available, so we should enable it and make this part of our QA process.

alexdewar:
As discussed.

I'm a bit swamped today, but don't want this to be a blocker for you, so if you're happy with it, just merge it. And if you need anything from me, message me on Teams.

Closes #128.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,3 @@
+BasedOnStyle: LLVM
+IndentWidth: 4
+ColumnLimit: 100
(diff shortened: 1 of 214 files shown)


jamesturner246 (approved):
Understood, thanks a lot @alexdewar. Just a couple of missing `#include`s exposed by the header reordering. All good now.

Issue #130, opened 2023-06-20
Merge risk-factor model classes and model definition classes.


jamesturner246:
Right now, `EnergyBalanceHierarchicalModel`, `StaticHierarchicalLinearModel`, etcetera all subclass `HierarchicalLinearModel`, but their respective definition classes `LiteHierarchicalModelDefinition`, `HierarchicalLinearModelDefinition` do not. This makes it messy when registering new models and model definitions with the risk factor module factory.

To simplify, we should merge the model definitions into their respective model classes.

Side note: consider renaming `HierarchicalLinearModel` to something more general, since these models are not necessarily HLMs any more. In fact, a lot of the above could be made more readable by renaming consistently.

Issue #107, opened 2023-05-16
Height variable


jamesturner246:
EDIT: It looks like this will finally be addressed. We will add height initialisation and update code to the `KevinHallModel` class.

These functions will depend on factor means, which are already loaded into the model. They will also depend on three extra variables, which must be added to the `KevinHall.json` config, and passed into the model ctor via the model parser. These variables are:

* Slope = 0.11,
* StdDevMale = 0.04,
* StdDevFemale = 0.04.

I have began work in the `height` branch, where I have implemented some stub methods as comments, and the `compute_mean_weight_powers` method, which will be used in common code used by the init and update height methods.

Code for this is available in the attached. It is not explained very well what is going on, or the variable names/units, but we can worry about explaining that after.

[Height_Regression.cs.pdf](https://github.com/imperialCHEPI/healthgps/files/13325883/Height_Regression.cs.pdf)

NB: remove the `.pdf` extension -- GitHub won't let me upload without it.

***BELOW IS OUTDATED:***

>Height variable in risk factors. 
>We will have mean heights for M/F by age from the config json (in  progress).

>- [ ] Add `height` to `Person` struct (with code to access as a risk factor)
>- [ ] initialise (see Ali's code: `Height.cs`)
>- [ ] update (see Ali's code `Height.cs`)
>- [ ] Dynamic height model as an LMS mdoel in `DemographicsModule`

jamesturner246:
height progression:

At first, we could just get height from age in a static lookup table. This can be updated in Person by the Demographics module?

Hardcoded average lookup, or taken from data in a country (e.g. India)?

Can eventually look at stochasticity -- maybe not worth a full model for height right now.

jamesturner246:
A hypothetical dynamic `height` update model inside `DemographicModule` would probably look something like the risk-factor-adjusting `LmsModel`, inside `WeightModel`, inside `DiseaseModule`.

This would be called inside `update_population`, and take a `Person`, with it's fields such as `age`, `sex`, `height`, and update `height` accordingly.

jamesturner246:
Discussion up to this point is outdated.

PR #144, opened 2023-07-03
New EnergyBalanceModel variables

### LINKED ISSUE #140: Add EBM equation variable fields to dynamic model config json
- [x] Add fields for equation variables to `./example/France.NewEBM.json`.
- [x] Add code to `model_parser` to read them.

For this change, we will add dummy energy intake coefficients for nutrient groups to `./example/France.NewEBM.json`:

```json
    "Nutrients": [
        {
            "Name": "Carbs",
            "Energy": 4.0
        },
        ...
    ],
```

We also will add average height at ages 0 to 100 data for each gender from the attached provided data file: [HeightData.csv](https://github.com/imperialCHEPI/healthgps/files/11916718/HeightData.csv)

```json
    "AgeMeanHeight": {
        "Male": [
            0.498, 0.757, 0.868, 0.952, 1.023, 1.092, 1.155, 1.219, 1.28, 1.333,
            ...
        ],
        "Female": [
            0.492, 0.741, 0.855, 0.94, 1.003, 1.079, 1.155, 1.211, 1.282, 1.333,
            ...
        ]
    }
```

We also need code to load this data (in `parse_model.cpp` into the new `EnergyBalanceModelDefinition` and `EnergyBalanceModel` classes.

jamesturner246:
NOTE: #130 should be merged before this.

This change introduces energy intake coefficients for each nutrient group in a NewEBM config JSON file. It also introduces a means of inputting the mean height value at any age for both male and female groups.

Nutrient energy coefficients are now specified by the following in `./example/France.NewEBM.json`, as opposed to the old `Nutrients` field:

```json
    "Nutrients": [
        {
            "Name": "Carbs",
            "Energy": 4.0
        },
```

and mean height at age are specified with:

```json
    "AgeMeanHeight": {
        "Male": [
            0.498, 0.757, 0.868, 0.952, 1.023, 1.092, 1.155, 1.219, 1.28, 1.333,
            ...
```

These values are loaded into the `EnergybalanceModelDefinition` via `model_parser.cpp`, and energy intake is computed in `EnergyBalanceModel`'s `update_risk_factors` method. Dummy values are used (we need to model height per-person once #107 is done.

Fixes #140


example/France.NewEBM.json
```diff
@@ -33,21 +50,31 @@
...
-    "RiskFactors": [
-        {
-            "COMMENT": "do we even need to list risk factors here?",
-            "Name": "BMI",
-            "Level": 3,
-            "Proxy": "",
-            "Range": [13.88, 39.48983]
-        }
-    ]
+    "AgeMeanHeight": {
```
dalonsoa, line 53: What are the corresponding ages? Is it always assume that it is one point per year? Not sure of the implementation details, but it feels very brittle.

src/HealthGPS.Console/model_parser.cpp
```diff
@@ -174,43 +177,59 @@ LiteHierarchicalModelDefinition load_dynamic_risk_model_definition(const host::p
...
 
-    return EnergyBalanceModelDefinition(std::move(nutrient_list), std::move(nutrient_equations));
+    // Save M/F average heights for age.
+    auto male_height = opt["AgeMeanHeight"]["Male"].get<std::vector<double>>();
+    age_mean_height.emplace(core::Gender::male, std::move(male_height));
+    auto female_height = opt["AgeMeanHeight"]["Female"].get<std::vector<double>>();
+    age_mean_height.emplace(core::Gender::female, std::move(female_height));
+
+    return std::make_shared<EnergyBalanceModelDefinition>(
+        std::move(energy_equation), std::move(nutrient_equations), std::move(age_mean_height));
```
dalonsoa, line 223: As mentioned, this feels brittle. What if the male and female vectors (why vectors and not arrays, by the way?) have different length? What if they cover just 70 years instead of 100 years? What if they are defined every 2 years instead of every year? Even if this is properly documented, it can break very easily. 

I think there should be a sort of validation step here to make sure the data is consistent, as well as the requirement to include the years in the input file. If they need to be defined every year 0-100, then you can use this step to interpolate/extrapolate the input data.

src/HealthGPS.Console/model_parser.cpp
```diff
@@ -174,43 +177,59 @@ LiteHierarchicalModelDefinition load_dynamic_risk_model_definition(const host::p
...
+        for (const auto &nutrient : opt["Nutrients"]) {
+            auto nutrient_str = nutrient["Name"].get<std::string>();
+            auto nutrient_key = core::Identifier{nutrient_str};
 
-            if (food_nutrient_strings.contains(nutrient_str)) {
-                double val = food_nutrient_strings.at(nutrient_str);
-                nutrient_equations[food_ident][nutrient_ident] = val;
+            if (food_nutrients.contains(nutrient_str)) {
+                double val = food_nutrients.at(nutrient_str);
+                nutrient_equations[food_key][nutrient_key] = val;
```
dalonsoa, line 211: Why do the nutrients provided directly have a value of "Energy" but those provided via "Foods" do not?

src/HealthGPS/energy_balance_model.h
```diff
@@ -11,12 +11,18 @@ namespace hgps {
...
-    /// @param definition The model definition instance
-    EnergyBalanceModel(EnergyBalanceModelDefinition &definition);
+    /// @param energy_equation The energy coefficients for each nutrient
+    /// @param nutrient_equations The nutrient coefficients for each food group
+    /// @param age_mean_height The mean height at all ages (male and female)
+    EnergyBalanceModel(
+        const std::unordered_map<core::Identifier, double> &energy_equation,
+        const std::unordered_map<core::Identifier, std::map<core::Identifier, double>>
+            &nutrient_equations,
+        const std::unordered_map<core::Gender, std::vector<double>> &age_mean_height);
```
dalonsoa, line 21: Personally, I think we should be talking about `sex` in these equations and not `gender`, which is a different thing, but I'm not going to be the one telling them what to use...

dalonsoa (changes requested):
My main comment is in reference to handling the input file. Otherwise, it looks good - as far as I can tell :)

src/HealthGPS/energy_balance_model.h
jamesturner246: Fair point. I suppose there won't be complaints if they don't see the change!

src/HealthGPS.Console/model_parser.cpp
jamesturner246: I plan to do a big input sanitisation PR soon, but not against adding the checks now. Why vectors and not arrays? Arrays would indeed be more appropriate here. Force of habit.

example/France.NewEBM.json
jamesturner246: In this case, age 0 is the 0th index, etcetera. It assumes age data is complete from zero to max age, which *should* be the case -- no point leaving gaps. Maybe this is a bad assumption?

dalonsoa: I have no idea if that's a bad assumption - it depends what the real data usually looks like. If you're asked to provide the years, then there's no ambiguity.

src/HealthGPS.Console/model_parser.cpp
jamesturner246: Ah, it turns out to use `std::array`s, the container size must be be known at compile time (array arguments for functions must be of the form `void fn (std::array<int, ARRAY_SIZE> a) {}`. So we need `std::vector`s here because we don't know upfront how many ages (thus array size) are needed.

example/France.NewEBM.json
jamesturner246: I think the solution here is proper input sanitisation, making sure that ages are defined and complete up to a minimum of max person age defined in the main config file. Missing data should be a bug.

src/HealthGPS.Console/model_parser.cpp
```diff
@@ -184,32 +184,43 @@ load_dynamic_risk_model_definition(const host::poco::json &opt) {
...
+        auto nutrient_energy = nutrient["Energy"].get<double>();
+        energy_equation[nutrient_key] = nutrient_energy;
     }
 
     // Save food -> nutrient equations.
     for (const auto &food : opt["Foods"]) {
-        auto food_ident = core::Identifier{food["Name"].get<std::string>()};
-        auto food_nutrient_strings = food["Nutrients"].get<std::map<std::string, double>>();
+        auto food_str = food["Name"].get<std::string>();
+        auto food_key = core::Identifier{food_str};
```
alexdewar, line 202: Again, `food_str` could be moved.

As this seems to be a fairly common pattern in this file, how about making a helper function, called `get_identifier()` or something? It would take in the JSON object and return a `core::Identifier`.

alexdewar (changes requested):
Looks good!

I've made a few comments and a few suggestions about ``std::move``ing things (as usual), but otherwise all good :smile:.

I'd open an issue about checking the input data if you haven't already, as I'm a little concerned that the existing code (not yours in particular!) is rather bug-prone in this respect. It might be worth implementing this sooner rather than later, to make it easier to debug things when you're trying to modify the input file format.

src/HealthGPS.Console/model_parser.cpp
alexdewar: Actually, I've got a better suggestion.

The JSON library lets you serialise/deserialise arbitrary types if you provide helpers: https://json.nlohmann.me/features/arbitrary_types/#basic-usage

So you should be able to just put this little shim somewhere:

```cpp
namespace hgps::core {
void from_json(const host::poco::json& j, Identifier& id) {
    id = Identifier(j.get<std::string>());
}
} // namespace hgps::core
```

Then you can do:

```cpp
auto food_key = food["Name"].get<core::Identifier>();
```

example/France.NewEBM.json
jamesturner246: I will do sanitation of the variables above in this PR. I think a wider sanitisation effort would need its own PR though.

src/HealthGPS.Console/model_parser.cpp
jamesturner246: Ah, good shout. Forgot about that.

src/HealthGPS/energy_balance_model.h
jamesturner246: Maybe leave this on the back burner then. I'll open an issue anyway.

src/HealthGPS.Core/identifier.cpp
```diff
@@ -55,4 +55,13 @@ std::ostream &operator<<(std::ostream &stream, const Identifier &identifier) {
...
     return stream;
 }
-} // namespace hgps::core
\ No newline at end of file
+
+void from_json(const nlohmann::json &j, Identifier &id) { id = Identifier{j.get<std::string>()}; }
+
+void from_json(const nlohmann::json &j, std::map<Identifier, double> &map) {
+    for (auto &[key, value] : j.items()) {
+        map.emplace(Identifier{key}, value.get<double>());
```
alexdewar, line 63: I *think* you can just construct the `Identifier` in place here:

```suggestion
        map.emplace(key, value.get<double>());
```

alexdewar (approved):
I've made one final suggestion, but otherwise LGTM!

PR #145, opened 2023-07-05
Fix documentation generation on CI

### LINKED ISSUE #131: Doxygen documentation generation is broken
PR #117 unfortunately broke the doxygen documentation generation, so all the API documentation has been deleted.

Looks like I need to do some more config file wrangling.

alexdewar:
It turned out to be a one-line config file fix. I forgot to update one of the places where a variable was used when I merged the CI configs.

I've tested this by commenting out the `if` guard that makes documentation only build for releases and new documentation was pushed to `gh-pages`.

Fixes #131.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -94,7 +94,7 @@ jobs:
         uses: peaceiris/actions-gh-pages@v3
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


jamesturner246 (approved):
Fair enough. All the CI stuff runs on Linux anyway, right? Go ahead.

alexdewar:
Nope, the CI also runs on Windows, but only for building Windows binaries. You're right that the documentation and pre-commit stuff only runs on Linux.

PR #146, opened 2023-07-05
Fix: Don't unconditionally catch exceptions in `main()`

### LINKED ISSUE #121: Don't unconditionally catch exceptions in `main()`
The problem is that if exceptions are always caught, then the debugger can't intercept them. Among other things, this also means you don't get a backtrace, which limits the utility of the printed message.

Part of the issue is that the default handler for uncaught exceptions on Windows just gives a dialog saying that an error has occurred, without giving more information. So we do want to continue to print out the exception message, at least on Windows, but we do also ideally then want to rethrow the exception so the debugger can intercept it. We could restrict the rethrowing to debug builds. See example: https://github.com/BrainsOnBoard/bob_robotics/blob/5e525c0e8603f55c2fef6fcec5f84b6903ebca50/src/common/main.cc#L187

alexdewar:
Currently, any uncaught exceptions in Health-GPS will percolate up to the `main()` function, where there is a `try...catch` block around all the code. The `catch` part just displays an error message (minus the stack trace!) then exits the program. This makes debugging difficult as debuggers will not be able to break into the code where the exception was thrown or display a stack trace.

The only reason for keeping the `try...catch` and not getting rid of it altogether is that on Windows, the default handler for uncaught exceptions just displays an uninformative "error has occurred" dialog to users, without giving any more information.

In this PR I've made this behaviour configurable at compile time. By default, exceptions will be caught on release builds and not on debug builds, however I have added a `CATCH_EXCEPTIONS` CMake option in case the user wants to opt in/out of this (e.g. if debugging a release build).

Closes #121.

### DIFF
--- a/src/HealthGPS.Console/program.cpp
+++ b/src/HealthGPS.Console/program.cpp
@@ -52,7 +52,9 @@ int main(int argc, char *argv[]) {
         core::run_async(load_datatable_from_csv, std::ref(input_table), config.file.name,
                         config.file.columns, config.file.delimiter);
 
+#ifdef NDEBUG
     try {
+#endif
         // Create back-end data store, cached data repository wrapper
         auto data_api = data::DataManager(cmd_args.storage_folder, config.verbosity);
         auto data_repository = hgps::CachedRepository{data_api};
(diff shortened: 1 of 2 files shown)


src/HealthGPS.Console/program.cpp
```diff
@@ -156,9 +158,15 @@ int main(int argc, char *argv[]) {
         std::this_thread::sleep_for(std::chrono::milliseconds(200));
         fmt::print(fg(fmt::color::light_green), "\nCompleted, elapsed time : {}ms\n\n", runtime);
         event_monitor.stop();
+
+#ifdef CATCH_EXCEPTIONS
     } catch (const std::exception &ex) {
         fmt::print(fg(fmt::color::red), "\n\nFailed with message: {}.\n\n", ex.what());
+
+        // Rethrow exception so it can be handled by OS's default handler
+        throw;
```
jamesturner246, line 167: Is it better to call `std::rethrow_exception (ex)` or just throw in this case. Not a nitpick, genuine question!

jamesturner246 (approved):
Okay, thanks. LGTM.

PR #148, opened 2023-07-05
CI: Make compiler warnings fatal

alexdewar:


CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 cmake_minimum_required (VERSION 3.20)
 
-option(BUILD_BENCHMARKS "Build benchmark application" OFF)
 option(BUILD_DOC "Build API documentation" ON)
+option(WARNINGS_AS_ERRORS "Make the build fail if the compiler gives a warning" OFF)
```
dalonsoa, line 4: So, let's see if I understand this: when running things locally, the warning will not be considered errors (therefore this `OFF`), but when running in the CI, they will be errors (therefore the `ON` in the workflow file), right?

Why not considering warnings as errors locally, as well?

src/HealthGPS.Benchmarks/BenchMain.cpp
dalonsoa, line 1: What was this benchmark doing?

dalonsoa (approved):
Looks good! Just a couple of questions.

src/HealthGPS.Benchmarks/BenchMain.cpp
jamesturner246: Just testing string hashing versus hashing the strange custom `core::Identifier` strings for `std::map` addressing. As @alexdewar said, not super useful to have now.

jamesturner246 (approved):

CMakeLists.txt
alexdewar: Yep! If it's turned on the compiler will exit with a code other than zero. On GitHub Actions that means the job will fail, which is what we want, because we want to enforce warning-free builds.

The problem is that different compilers give different warnings and under different circumstances (e.g. they sometimes differ in whether they're able to determine that a given variable is unused), so if you turn the option on by default it means that if someone's using, say, a newer version of gcc than we are is likely to have the build fail, because of new warnings that have been added, which isn't particularly user friendly if they aren't one of the Health-GPS developers. There's a good explanation of this rationale here: https://embeddedartistry.com/blog/2017/05/22/werror-is-not-your-friend/

We should also add other compilers to the CI to improve our coverage though, especially clang and whatever the latest version of gcc is at the time.

alexdewar:
@jamesturner246 You can turn it on locally if you want when you configure the project:

```sh
cmake --preset linux-debug -DWARNINGS_AS_ERRORS=ON
```

But as I said to @dalonsoa, it's quite annoying for non-developers if you leave it on by default.

PR #150, opened 2023-07-11
Remove unnecessary bounds checks

### LINKED ISSUE #123: Remove unnecessary bounds checks
There are a bunch of places in the code where the code checks whether an index is valid, before accessing it using the `at()` method. As `at()` already performs its own bounds checking, the initial check is redundant.

See, for example: https://github.com/imperialCHEPI/healthgps/blob/d53866cf85ad03397702dad6aa86408e6a1f40d3/src/HealthGPS/life_table.cpp#L36

alexdewar:
I did this last week and forgot to open a PR for it.

There weren't actually as many of these as I initially thought. I've removed them anyway.

I also spotted one argument that could be made into a const ref.

Closes #123.

### DIFF
--- a/src/HealthGPS.Core/datatable.cpp
+++ b/src/HealthGPS.Core/datatable.cpp
@@ -34,13 +34,7 @@ void DataTable::add(std::unique_ptr<DataTableColumn> column) {
     columns_.push_back(std::move(column));
 }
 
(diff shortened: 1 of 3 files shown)


dalonsoa (approved):
All good!

jamesturner246 (approved):
Less is more!

PR #153, opened 2023-07-11
Use `std::unique_ptr`s for storing model definitions 

alexdewar:
I fancied tinkering with some code rather than just config files, so I did this, even though it's not in the backlog and is a pretty minor refactor :stuck_out_tongue:.

Currently, the model definitions are stored in the repository (of type ``CachedRepository``) as ``std::shared_ptr``s, even though they don't really need to be "shared", because the repository object will outlive any of the users of these model definitions. So instead, I've changed these to be ``std::unique_ptr``s; when they're needed they can be accessed as good old-fashioned const refs.

I also changed a couple of small issues I spotted on the way.

### DIFF
--- a/src/HealthGPS.Console/model_parser.cpp
+++ b/src/HealthGPS.Console/model_parser.cpp
@@ -43,7 +43,7 @@ hgps::BaselineAdjustment load_baseline_adjustments(const poco::BaselineInfo &inf
     }
 }
 
-std::shared_ptr<hgps::HierarchicalLinearModelDefinition>
+std::unique_ptr<hgps::HierarchicalLinearModelDefinition>
 load_static_risk_model_definition(const poco::json &opt) {
     MEASURE_FUNCTION();
     std::map<int, hgps::HierarchicalLevel> levels;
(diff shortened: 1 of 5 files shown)


dalonsoa (commented):
James explained to me the difference between `shared_ptr` and `unique_ptr` and while I did not grasped everything, I gathered that using the second option was not possible in this case since we need to have multiple copies of it in different places. Needless to say that I'm not an expert in C++, so maybe I got this totally wrong....

alexdewar:
That sounds v bizarre 😆 but I'm sure there's a good reason for it! Yes, you could easily modify it to be non-`const`. The thing is though, if you do that it will also modify the version in the repository (and therefore any future models will also have modified params). Would that be a problem?

I did wonder if we might be able to remove the model definition classes, but having had a look just now, I'm wondering how we'd be able to construct models without having the params encapsulated in a separate class somewhere? (That's assuming that we actually *do* want to cache model parameters rather than reloading them every time.)

jamesturner246:
The only sensible reason I could think of for caching parameters was if they wanted some persistent interface in future, where they could stop, reinitialise, restart, etcetera, which would mean re-initialising from some pristine confifg. In which case, you're right, it's probably better to keep the 'definition' classes, keep the cache constant, and just copy out of it whenever they need fresh state.

If they don't want that, then there's no reason to cache anything. But I suppose they *might* want that in future, so I guess that's reason enough to keep the pristine cache.

dalonsoa:
Rather than wondering what they might want, why do you ask them (after giving some context of the issue)?

jamesturner246 (approved):

jamesturner246:
I'll bring it up with them this week.

jamesturner246:
Ali has just confirmed that future plans might include a GUI which would need a persistent pristine configuration cache to generate models from, so I would say that this is the correct approach, and to go ahead and merge this.

Issue #122, opened 2023-06-16
Consider dropping debug build from CI


alexdewar:
We don't obviously derive much benefit from it and it increases the time the CI takes to run.

Issue #73, opened 2022-11-04
[OSB] Missing ``#include`` in ``src/HealthGPS.Core/column_primitive.h``


alexdewar:
It fails to compile on gcc 12 unless ``#include <algorithm>`` is added, as the ``std::count()`` function is used later on.

PR #154, opened 2023-07-11
Add clang and latest gcc to CI system

### LINKED ISSUE #76: [OSB] CI: Also check that code compiles cleanly with clang
While it is always nice for a project to support use of more compilers, the main purpose of supporting clang is *not* to support it for its own sake. The main purpose is because clang often identifies mistakes in code that gcc doesn't (see #74 and #75), so by adding it to your list of targets in the CI system you can catch bugs earlier on in the development process; you're essentially using it as a linter.

Clang was designed as a gcc clone so it has a very similar interface and, apart from the minor issues I've linked to, adding support should be very easy.

### LINKED ISSUE #84: [OSB] Suggestion: Also check that code compiles with latest gcc on CI
At the time of writing the latest version of gcc is v12, while the CI tests are running on v11. I would suggest also testing on v12.

While I'm not saying that you necessarily need to always be using the latest version of gcc, it makes sense to at least test that your code compiles with it because:

1. Your users might be using it
2. You'll probably end up using it at some point, so it is useful to know about problems sooner rather than later
3. New versions of gcc sometimes introduce new warnings that can identify problems with your code

Often the changes required to support a new version of gcc are pretty minimal (I only found one build issue: #73), so there shouldn't be a great cost to doing this.

alexdewar:
This PR adds clang and the latest version of gcc to the compilers tested on the CI (both running on Ubuntu), which should increase our coverage.

Closes #76. Closes #84.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,12 +23,16 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        platform: [linux, windows]
         include:
           - platform: linux
             os: ubuntu-22.04
+            compiler: gcc
+          - platform: linux
+            os: ubuntu-22.04
+            compiler: clang
           - platform: windows
             os: windows-2022
+            compiler: msvc
 
     env:
       VCPKG_ROOT: "vcpkg"
@@ -50,6 +54,14 @@ jobs:
           sudo apt upgrade -y
           sudo apt install --no-install-recommends build-essential libtbb-dev ninja-build doxygen graphviz
 
+      - name: Install clang (Linux)
+        if: matrix.platform == 'linux' && matrix.compiler == 'clang'
+        run: |
+          sudo apt install --no-install-recommends clang
+
+          # Use environment variable to tell CMake to compile with clang
+          echo "CXX=clang++" >> "$GITHUB_ENV"
+
       - name: Install VCPKG
         run: |
           git clone https://github.com/Microsoft/vcpkg.git
@@ -69,6 +81,7 @@ jobs:
           cmake --build --preset=release-build-${{ matrix.platform }} --target=install
 
       - name: Upload artifacts
+        if: matrix.compiler != 'clang'
         uses: actions/upload-artifact@v3
         with:
           name: health-gps-${{ matrix.platform }}
@@ -79,20 +92,20 @@ jobs:
         run: ctest --preset=core-test-${{ matrix.platform }}
 
       - name: Zip output folder
-        if: startsWith(github.ref, 'refs/tags/')
+        if: startsWith(github.ref, 'refs/tags/') && matrix.compiler != 'clang'
         working-directory: ${{github.workspace}}/out/install/${{ matrix.platform }}-release/bin/
         run: |
           mkdir ${{github.workspace}}/artifact
           7z a -tzip ${{github.workspace}}/artifact/health_gps_${{ matrix.platform }}.zip *.dll *.Console*
 
       - name: Upload release artifacts
-        if: startsWith(github.ref, 'refs/tags/')
+        if: startsWith(github.ref, 'refs/tags/') && matrix.compiler != 'clang'
         uses: softprops/action-gh-release@v1
         with:
           files: artifact/health_gps_${{ matrix.platform }}.zip
 
       - name: Doxygen release API deploy
-        if: matrix.platform == 'linux' && startsWith(github.ref, 'refs/tags/')
+        if: matrix.platform == 'linux' && matrix.compiler != 'clang' && startsWith(github.ref, 'refs/tags/')
         uses: peaceiris/actions-gh-pages@v3
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,12 +23,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-22.04, windows-2022]
+        platform: [linux, windows]
         include:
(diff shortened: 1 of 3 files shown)


src/HealthGPS/energy_balance_model.cpp
```diff
@@ -51,7 +51,7 @@ void EnergyBalanceModel::update_risk_factors(RuntimeContext &context) {
             current_risk_factors.at(age_key) = model_age;
         }
 
-        double energy_intake = 0.0;
+        [[maybe_unused]] double energy_intake = 0.0;
```
jamesturner246, line 54: I assume the new linters pick this kind of stuff out automatically now, right?

.github/workflows/ci.yml
```diff
@@ -79,20 +108,20 @@ jobs:
         run: ctest --preset=core-test-${{ matrix.platform }}
 
       - name: Zip output folder
-        if: startsWith(github.ref, 'refs/tags/')
+        if: startsWith(github.ref, 'refs/tags/') && (matrix.platform != 'linux' || matrix.compiler == 'gcc-default')
```
jamesturner246, line 111: Should it be `matrix.platform == 'linux'`, or am I being dumb? I'm not sure I completely understand these lines.

jamesturner246 (approved):
Approve in principle. Just a couple of questions.

src/HealthGPS/energy_balance_model.cpp
alexdewar: I didn't add new linters -- just new compilers -- but, yes, clang noticed that this variable was unused even though gcc didn't.

.github/workflows/ci.yml
alexdewar: It's not obvious... I should probably add a comment.

This step and the one after it are only run for new releases. They zip up the binaries then upload them to GitHub as release artifacts. We want to run this step for every platform, but we only want to upload one set of binaries for Linux, even though we're building it several times with different compilers.

.github/workflows/ci.yml
```diff
@@ -23,33 +23,61 @@ jobs:
...
-          - os: ubuntu-22.04
-            platform: linux
-          - os: windows-2022
-            platform: windows
+          - platform: linux
+            os: ubuntu-22.04
+            compiler: gcc-default
+          - platform: linux
+            os: ubuntu-22.04
+            compiler: gcc-latest
```
dalonsoa, line 32: C++ is supposed to be backwards compatible, so if things work with an older gcc version, they should work with the latest gcc version, right? In that sense, even though I understand this might come with some extra warnings and niceties, I'm not totally sure it is worth having another branch for it - at least not with every push.

I'm not sure how much you got in the standup yesterday about being environmental-aware when setting our CI systems (your connection was a bit unstable), so we don't waste resources and energy unnecessary.

.github/workflows/ci.yml
dalonsoa: I think that mixing platforms and compiles muddy the explanation a bit. Maybe use, so you choose what to upload based on the compiler only.

```suggestion
        if: startsWith(github.ref, 'refs/tags/') && (matrix.compiler == 'msvc' || matrix.compiler == 'gcc-default')
```

dalonsoa (approved):
I've made a couple of comments, but looks good otherwsie.

.github/workflows/ci.yml
alexdewar: I didn't hear much so I was left wondering what you'd said :laughing:.

Yes, C++ is in theory backwards compatible. In practice though, you do often get compile errors when you move to a new compiler, usually because one header in your compiler's implementation of the standard library used to `#include` another and now it doesn't (e.g. #73). Compilers do often vary in how much of the various standards they implement and how faithfully too, which can also cause breakage (I just learned that the latest version of Apple's clang doesn't implement STL parallel algorithms *at all*, even though that's a C++17 feature, which means you can't compile Health-GPS with it). That's why I think it makes sense to have some minimum version of `gcc` that you test -- we're using Ubuntu LTS for the runner, so the default version included with that seems a sensible candidate -- but also to test against the latest version. That way we can be fairly confident that anyone using any of the intervening versions will be alright too. The extra warnings are a nice bonus, but also the Ubuntu folks tend to ship fairly old versions of packages, including `gcc`, so I do think it makes sense to test against the latest version of `gcc` too for people who use other distros with a less ancient version. (@jamesturner246 uses Fedora and I use Arch, which both ship newer versions than you find in Ubuntu LTS.) It also means that we don't have to fix a bunch of warnings and errors whenever we update the version of the Ubuntu runner. Admittedly, that's not a big job, but it would be better to fix the problems in advance, both for people using a newer version of `gcc` and also because compiler warnings can indicate things like serious memory corruption bugs caused by undefined behaviour -- unfortunately there are lots of ways you can shoot yourself in the foot with C++.

We are currently only running the CI for pushes to `main` and for PRs, so hopefully the resource usage isn't too terrible. The other thing we could do is restrict ourselves to building a release version of the code, not the debug version (see #122).

.github/workflows/ci.yml
alexdewar: This would work now, but won't work when I add macOS to the runners. I agree this is a bit ugly though. How about I add another parameter which indicates whether the compiler is the "main" one or not (e.g. a `matrix.is_main` boolean value you could check instead)?

dalonsoa: Sounds like a good option

.github/workflows/ci.yml
dalonsoa: What a nice explanation - worth a blogpost ;) . I hadn't realised that this runs only in pushes to main. Ok, no comments, then.

PR #157, opened 2023-07-13
Ebm equations

### LINKED ISSUE #141: Implement EBM equations using fixed dummy data
Implement the EBM model equations in the `update_risk_factors` function in `EnergyBalanceModel`.

Port from Ali's C# code:  [Kevin Hall EBM.zip](https://github.com/imperialCHEPI/healthgps/files/11916946/Kevin.Hall.EBM.zip)

Depends on #140

The relevant code in the C# code zip is the `Population::ProjecrtPopulationIntoFuture` function, which first updates a few risk factor variables `Height`, `Carbs`, `Fat`, `Protein`, `Alcohol`, `PAL`, and computes new energy intake from nutrient -> energy coefficients, computed from food groups with food -> nutrient coefficients.

energy and nutrient coefficients will be respectively made available in `EnergyBalanceModel.energy_equation_` and `EnergyBalanceModel.nutrient_equations_` from #140 and earlier. `Carbs`, `Fat`, `Protein`, `Alcohol`, `PAL` can be constant dummy values for now, as they depend on missing data, while `Height` is dependant on #107, so can be a dummy value meanwhile.

jamesturner246:
Okay. Finally. It's a bit rough around the edges and hacky, but that's probably close to the best we can get until proper data comes in.

The main method in `EnergyBalanceModel` is `update_risk_factors`, which calls `simulate_person` twice per year: once for a trial run to determine if the simulation diverges from statistical norms for a population (and computes a baseline adjustment), and another 'final' run, in which the baseline adjustment is applied.

Currently data is read from a very hacky construct in `Person`, which just returns dummy values.  Shield your eyes when you find it. -_-

There's very likely a nicer way to handle the return value, trial or final flag and offset argument of `simulate_person`. I haven't thought of it yet, and just wanted to get the PR going for now.

Fixes #141


src/HealthGPS/energy_balance_model.cpp
```diff
@@ -35,66 +46,163 @@ void EnergyBalanceModel::generate_risk_factors([[maybe_unused]] RuntimeContext &
...
+    double mean_sim_body_weight = 0.0;
+    double mean_adjustment_coefficient = 0.0;
+
+    // TODO: Compute target body weight.
+    const float target_BW = 100.0;
+
+    // Trial run.
+    for (auto &person : population) {
+        // Ignore if inactive.
+        if (!person.is_active()) {
```
alexdewar, line 60: This isn't a big deal, but if you *really* care about performance, iterating over a big `std::vector` and only doing something if a flag is (un)set is not the most efficient way to do it, because your CPU is constantly branching. It's faster if you have two separate vectors (in this case, one for active and one for inactive) and you just iterate over the one you're interested in.

As I said, def not important for now, but this might be something to look at if we're trying to optimise in future. (The fact that each person is a separate object is also a red flag from a performance standpoint!)

src/HealthGPS/energy_balance_model.cpp
```diff
@@ -35,66 +46,163 @@ void EnergyBalanceModel::generate_risk_factors([[maybe_unused]] RuntimeContext &
...
-            get_current_risk_factors(context.mapping(), entity, context.time_now());
-
-        // Model calibrated on previous year's age
-        auto model_age = static_cast<int>(entity.age - 1);
-        if (current_risk_factors.at(age_key) > model_age) {
-            current_risk_factors.at(age_key) = model_age;
-        }
+        auto trial = simulate_person(person, false, 0.0);
+        mean_sim_body_weight += trial[0];
+        mean_adjustment_coefficient += trial[1];
```
alexdewar, line 66: I think you can do this:

```suggestion
        auto [body_weight, adjustment] = simulate_person(person, false, 0.0);
        mean_sim_body_weight += body_weight;
        mean_adjustment_coefficient += adjustment;
```

Modern C++ ftw.

src/HealthGPS/energy_balance_model.cpp
```diff
@@ -35,66 +46,163 @@ void EnergyBalanceModel::generate_risk_factors([[maybe_unused]] RuntimeContext &
...
+        simulate_person(person, true, final_shift);
+    }
+}
 
-            for (const auto &coefficient : equation.second) {
-                double delta_nutrient = food_intake * coefficient.second;
-                nutrient_intakes[coefficient.first] += delta_nutrient;
-            }
-        }
+std::array<double, 2> EnergyBalanceModel::simulate_person(Person &person, bool final_run,
```
alexdewar, line 85: I feel like this should return a `std::pair<double, double>` because the values are unrelated.

src/HealthGPS/energy_balance_model.cpp
```diff
@@ -35,66 +46,163 @@ void EnergyBalanceModel::generate_risk_factors([[maybe_unused]] RuntimeContext &
...
+    double H = H_0;
+
+    // TODO: Update physical activity level.
+    double PAL = PAL_0;
+
+    // Compute nutrient intakes from food intakes.
+    std::unordered_map<core::Identifier, double> nutrient_intakes;
+    for (const auto &coefficient : energy_equation_) {
+        nutrient_intakes[coefficient.first] = 0.0;
+    }
```
alexdewar, line 108: STL containers default-initialise their contents anyway, so you don't need this:

```suggestion
```

src/HealthGPS/energy_balance_model.cpp
```diff
@@ -35,66 +46,163 @@ void EnergyBalanceModel::generate_risk_factors([[maybe_unused]] RuntimeContext &
...
+    const double p = C / (C + F_0);
+
+    // First equation ax + by = e.
+    double a1 = p * rho_F;
+    double b1 = -(1.0 - p) * rho_L;
+    double c1 = p * rho_F * F_0 - (1 - p) * rho_L * L_0;
+
+    // Check no baseline adjustment in trial run.
+    if (!final_run)
+        final_shift = 0.0;
```
alexdewar, line 162: Is this something that could be enforced with an exception?

src/HealthGPS/energy_balance_model.cpp
```diff
@@ -35,66 +46,163 @@ void EnergyBalanceModel::generate_risk_factors([[maybe_unused]] RuntimeContext &
...
+        person.risk_factors[L_key] = L;
+        person.risk_factors[ECF_key] = ECF;
+        person.risk_factors[EI_key] = EI;
+        person.risk_factors[EE_key] = EE;
+        return std::array<double, 2>{};
+    } else { // Trial run:
+        // return BW and baseline adjustment coefficient.
+        double trial_adjust = -(a1 - b1) * (1.0 - exp(-365.0 / tau)) / (a1 * b2 - a2 * b1);
+        return std::array<double, 2>{BW, trial_adjust};
+    }
```
alexdewar, line 205: I'd drop the `else`. Also you don't need to specify the type here because the compiler can infer it:

```suggestion
        return {};
    }

    // Trial run:
    // return BW and baseline adjustment coefficient.
    double trial_adjust = -(a1 - b1) * (1.0 - exp(-365.0 / tau)) / (a1 * b2 - a2 * b1);
    return {BW, trial_adjust};
```

src/HealthGPS/healthgps.cpp
```diff
@@ -331,8 +332,13 @@ std::map<std::string, core::UnivariateSummary> HealthGPS::create_input_data_summ
     auto summary = std::map<std::string, core::UnivariateSummary>();
     auto &input_data = definition_.inputs().data();
     for (const auto &entry : context_.mapping()) {
-        input_data.column(entry.name()).accept(visitor);
-        summary.emplace(entry.name(), visitor.get_summary());
+        try {
+            input_data.column(entry.name()).accept(visitor);
+            summary.emplace(entry.name(), visitor.get_summary());
+        } catch (const std::out_of_range &oor) {
```
alexdewar, line 338: Unused variable?

src/HealthGPS/healthgps.cpp
```diff
@@ -331,8 +332,13 @@ std::map<std::string, core::UnivariateSummary> HealthGPS::create_input_data_summ
...
+            // HACK: ignore missing columns
+            continue;
```
alexdewar, line 340: This will be slow... Could you just `continue` if the column doesn't exist?

src/HealthGPS/person.cpp
```diff
@@ -30,34 +41,24 @@ unsigned int Person::time_of_migration() const noexcept { return time_of_migrati
...
-    }
-
-    // Dynamic properties
-    if (risk_factors.contains(key)) {
+    } else if (risk_factors.contains(key)) {
+        // Dynamic properties
         return risk_factors.at(key);
+    } else {
+        throw std::out_of_range("Risk factor not found: " + key.to_string());
     }
```
alexdewar, line 53: This is more of a personal thing, but I'd generally write something like this as:

```cpp
if (first_thing) {
    return 1;
}
if (second_thing) {
    return 2;
}

throw std::exception("bad thing happened!");
```

I find the cognitive load is a bit less than when you have a bunch of `else`s.

src/HealthGPS/energy_balance_model.cpp
alexdewar: I'm quite suspicious about this `final_run` argument; it kind of feels like you're trying to squash two functions into one here (with two different return types).

Would it make sense to just return the processed values (H, BW, etc.) in a struct? It could then be fed into either a function which updates the `Person` or calculates the trial adjustment depending on what the user wants to do with it. What do you reckon?

src/HealthGPS/energy_balance_model.cpp
```diff
@@ -35,66 +46,163 @@ void EnergyBalanceModel::generate_risk_factors([[maybe_unused]] RuntimeContext &
...
+    // Update thermic effect of food and adaptive thermogenesis.
+    double TEF = beta_TEF * EI - EI_0;
+    double AT = beta_AT * EI - EI_0;
+
+    // TODO: Update intercept value.
+    const double K = 0.0; // TODO: Intercept value.
+
+    // Energy partitioning.
+    const double C = 10.4 * rho_L / rho_F;
+    const double p = C / (C + F_0);
```
alexdewar, line 153: I think all the equations and update steps should be made into separate functions for testability.

alexdewar (changes requested):
Looking good! I've made some minor suggestions, but it's exciting that it's nearly there :smile:.

I've got no idea about whether you've implemented the equations correctly though, obvs. How about splitting them out into separate functions and writing a few tests for them?

src/HealthGPS/energy_balance_model.cpp
jamesturner246: Right. AoS vs SoA. I haven't even touched the `Population` code there, but I suspect there's tons of optimisation in there.

src/HealthGPS/energy_balance_model.cpp
```diff
@@ -3,6 +3,17 @@
...
+// Rsik factor keys.
+const core::Identifier H_key{"height"};
+const core::Identifier BW_key{"weight"};
+const core::Identifier PAL_key{"physical_activity_level"};
+const core::Identifier F_key{"body_fat"};
+const core::Identifier L_key{"lean_tissue"};
+const core::Identifier ECF_key{"extracellular_fluid"};
+const core::Identifier EI_key{"energy_intake"};
+const core::Identifier EE_key{"energy_expenditure"};
+const core::Identifier CI_key{"carbohydrate"};
```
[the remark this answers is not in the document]
jamesturner246: What's the difference in this case? I thought `constexpr` was for things like float initialisation?

dalonsoa (changes requested):
I've little to add to this one except what already pointed out by Alex: the function that calculates the stuff is way, way too big. It should be split into smaller chunks that have some sort of meaning and can be tested individually. And if it is about a published paper, include in the dosctring a reference to the paper (and the Eq. numbers within).

src/HealthGPS/energy_balance_model.cpp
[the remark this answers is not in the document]
jamesturner246: `const` expressions can be evaluated at runtime, e.g. `const std::string{argv[0]}`, whereas `constexpr` forces it to be evaluated at compile time, latest, a bit more like a C macro with type safety.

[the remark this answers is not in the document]
dalonsoa: Is there any practical advantage?

src/HealthGPS/person.cpp
jamesturner246: Huh. Interesting little differences. I actually find the explicitness helpful. Not married to either way though, if it helps the reviewers.

src/HealthGPS/energy_balance_model.cpp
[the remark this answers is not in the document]
jamesturner246: Code hygiene? You're basically telling the compiler and reader that what you see in the source is what you get at runtime, and under no circumstances is the value to change. More generally, certain things like static array initialisation require a `constexpr` as the size parameter: `type someArray[myConstExpr];`, else the compiler can't determine how big a stack frame to allocate when entering the function. Besides that, just code hygiene.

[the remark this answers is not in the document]
dalonsoa: C++ is odd...

src/HealthGPS/energy_balance_model.cpp
jamesturner246: Fair enough. The baseline adjustment would be computed unnecessarily again on the final run, but that's not such a huge deal.

src/HealthGPS/energy_balance_model.cpp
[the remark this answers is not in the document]
jamesturner246: Yes, it is. ^^

src/HealthGPS/energy_balance_model.cpp
jamesturner246: If the comment above about returning a struct of results is implemented, we wouldn't need `final_run`, and the offset would be enforced by `final_shift` alone.

src/HealthGPS/healthgps.cpp
jamesturner246: Fixed in [e0d7b0a](https://github.com/imperialCHEPI/healthgps/pull/157/commits/e0d7b0ad1658d3f6d95b02ebe373c82e66d6c290)

src/HealthGPS/energy_balance_model.cpp
alexdewar: Good point @dalonsoa. I'm afraid I've already merged #158 but I'll do it for #160!

alexdewar: After thinking about this some more, I'm wondering if these "person IDs" actually need to be stored *anywhere*. If you're simulating a population of e.g. a million people, it's really odd to represent that with a vector containing the numbers one to a million. Probably we can get away with just keeping track of the various "counts" (i.e., number of active people, dead people and emigrants).

A job for another day, though.

src/HealthGPS/energy_balance_model.cpp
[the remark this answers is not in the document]
alexdewar: @jamesturner246 If it isn't simple, then don't bother. As you say, not much difference here.

The nice thing about `constexpr` variables in general cf. `const` ones is that you can use them in contexts where you *need* to know the value at compile time, e.g.:

```cpp
constexpr auto i = 4;
std::array<int, i> my_array;
```

This won't work if `i` was just `const`.

The other thing is that you can make functions `constexpr` aware and then you can have basically arbitrary calculations happen at compile time rather than runtime.

src/HealthGPS/person.cpp
alexdewar: Totally up to you. This one is mostly a matter of taste, I think.

src/HealthGPS/energy_balance_model.cpp
alexdewar: Is there a way the baseline adjustment could be calculated in its own function?

Maybe once you've split this function up a bit, the way forward will seem more obvious.

jamesturner246: I've thought about it, but it's not obvious how to do this. It relies on mean simulated weights and mean of these adjustment coefficients for the whole population, so I'd have to sit down and do a bit of math to be able to decouple it.

src/HealthGPS.Core/datatable.cpp
```diff
@@ -46,6 +46,17 @@ const DataTableColumn &DataTable::column(const std::string &name) const {
...
+const std::pair<bool, DataTableColumn &>
+DataTable::column_if_exists(const std::string &name) const {
+    auto found = index_.find(to_lower(name));
+    if (found != index_.end()) {
+        return {true, *columns_.at(found->second)};
+    }
+
+    // Not found.
+    return {false, **columns_.end()};
+}
```
alexdewar, line 58: I think the `std::pair` could be replaced with a `std::optional`.

src/HealthGPS/energy_balance_model.cpp
```diff
@@ -78,131 +82,169 @@ void EnergyBalanceModel::update_risk_factors(RuntimeContext &context) {
...
+        // person.risk_factors[BW_key] = state.BW;
+        // person.risk_factors[PAL_key] = state.PAL;
+        // person.risk_factors[RMR_key] = state.RMR;
+        // person.risk_factors[F_key] = state.F;
+        // person.risk_factors[L_key] = state.L;
+        // person.risk_factors[ECF_key] = state.ECF;
+        // person.risk_factors[G_key] = state.G;
+        // person.risk_factors[W_key] = state.W;
+        // person.risk_factors[EE_key] = state.EE;
+        // person.risk_factors[EI_key] = state.EI;
```
alexdewar, line 98: I'm guessing you left this as a placeholder? I think it would be better to have a dedicated function for updating the risk factors from your `state` struct. It *could* potentially be a member method of `Person`.

src/HealthGPS/energy_balance_model.cpp
```diff
@@ -78,131 +82,169 @@ void EnergyBalanceModel::update_risk_factors(RuntimeContext &context) {
...
+    // Initial simulated person state.
     const double H_0 = person.get_risk_factor_value(H_key);
     const double BW_0 = person.get_risk_factor_value(BW_key);
     const double PAL_0 = person.get_risk_factor_value(PAL_key);
     const double F_0 = person.get_risk_factor_value(F_key);
     const double L_0 = person.get_risk_factor_value(L_key);
     const double ECF_0 = person.get_risk_factor_value(ECF_key);
+    const double G_0 = person.get_risk_factor_value(G_key);
     const double EI_0 = person.get_risk_factor_value(EI_key);
     const double CI_0 = person.get_risk_factor_value(CI_key);
```
alexdewar, line 112: In a similar vein to my comment above, maybe `Person` needs a `get_risk_factor_values()` method which returns a `SimulatePersonState`. Would that make sense?

src/HealthGPS/energy_balance_model.cpp
```diff
@@ -78,131 +82,169 @@ void EnergyBalanceModel::update_risk_factors(RuntimeContext &context) {
...
+                              .F = F,
+                              .L = L,
+                              .ECF = ECF,
+                              .G = G,
+                              .W = W,
+                              .EE = EE,
+                              .EI = EI,
+                              .adjust = adjust};
+
+    return state;
```
alexdewar, line 196: Don't commit this suggestion as `clang-format` will complain, but I'd do this in one line like so:

```suggestion
    // New simulated person state.
    return SimulatePersonState{.H = H,
                              .BW = BW,
                              .PAL = PAL,
                              .RMR = RMR,
                              .F = F,
                              .L = L,
                              .ECF = ECF,
                              .G = G,
                              .W = W,
                              .EE = EE,
                              .EI = EI,
                              .adjust = adjust};
```

src/HealthGPS/energy_balance_model.cpp
```diff
@@ -78,131 +82,169 @@ void EnergyBalanceModel::update_risk_factors(RuntimeContext &context) {
...
+}
+
+std::unordered_map<core::Identifier, double>
+EnergyBalanceModel::compute_nutrients(const Person &person) const {
+    std::unordered_map<core::Identifier, double> nutrient_intakes;
+
+    for (const auto &equation : nutrient_equations_) {
+        double food_intake = person.get_risk_factor_value(equation.first);
+
+        for (const auto &coefficient : equation.second) {
```
alexdewar, line 206: You can use structured bindings here to make this more readable, i.e.:

```cpp
for (const auto &[firstThing, secondThing] : nutrient_equations_) {
```

src/HealthGPS/energy_balance_model.cpp
```diff
@@ -78,131 +82,169 @@ void EnergyBalanceModel::update_risk_factors(RuntimeContext &context) {
...
-        person.risk_factors[L_key] = L;
-        person.risk_factors[ECF_key] = ECF;
-        person.risk_factors[EI_key] = EI;
-        person.risk_factors[EE_key] = EE;
-        return std::array<double, 2>{};
-    } else { // Trial run:
-        // return BW and baseline adjustment coefficient.
-        double trial_adjust = -(a1 - b1) * (1.0 - exp(-365.0 / tau)) / (a1 * b2 - a2 * b1);
-        return std::array<double, 2>{BW, trial_adjust};
+    for (const auto &coefficient : energy_equation_) {
```
alexdewar, line 219: Again, maybe use a structured binding here to make it more obvious what `coefficient.first` and `coefficient.second` refer to.

src/HealthGPS/energy_balance_model.cpp
```diff
@@ -78,131 +82,169 @@ void EnergyBalanceModel::update_risk_factors(RuntimeContext &context) {
...
+        EI += delta_energy;
     }
+
+    return EI;
+}
+
+double EnergyBalanceModel::compute_G(double CI, double CI_0, double G_0) const {
+    double k_G = CI_0 / (G_0 * G_0);
+    double G = sqrt(CI / k_G);
+    return G;
```
alexdewar, line 230:
```suggestion
    return sqrt(CI / k_G);
```

src/HealthGPS/energy_balance_model.cpp
```diff
@@ -78,131 +82,169 @@ void EnergyBalanceModel::update_risk_factors(RuntimeContext &context) {
...
+}
+
+double EnergyBalanceModel::compute_ECF(double EI, double EI_0, double CI, double CI_0,
+                                       double ECF_0) const {
+    double Na_b = 4000.0;
+    double Na_f = Na_b * EI / EI_0;
+    double Delta_Na_diet = Na_f - Na_b;
+    double ECF = ECF_0 + (Delta_Na_diet - xi_CI * (1.0 - CI / CI_0)) / xi_Na;
+    return ECF;
```
alexdewar, line 239:
```suggestion
    return ECF_0 + (Delta_Na_diet - xi_CI * (1.0 - CI / CI_0)) / xi_Na;
```

src/HealthGPS/energy_balance_model.cpp
```diff
@@ -78,131 +82,169 @@ void EnergyBalanceModel::update_risk_factors(RuntimeContext &context) {
...
+}
+
+double EnergyBalanceModel::compute_RMR(double BW, double H, unsigned int age,
+                                       core::Gender gender) const {
+    double RMR = 9.99 * BW + 6.25 * H * 100.0 - 4.92 * age;
+    RMR += gender == core::Gender::male ? 5.0 : -161.0;
+    RMR *= 4.184; // kcal to kJ
+    return RMR;
```
alexdewar, line 247: In case you're interested, I've used this library before for unit conversions in C++: https://github.com/nholthaus/units

It gives you compile-time dimensional analysis, which is quite neat. Maybe worth considering?

src/HealthGPS/healthgps.cpp
```diff
@@ -331,13 +331,13 @@ std::map<std::string, core::UnivariateSummary> HealthGPS::create_input_data_summ
...
     auto summary = std::map<std::string, core::UnivariateSummary>();
     auto &input_data = definition_.inputs().data();
+
     for (const auto &entry : context_.mapping()) {
-        try {
-            input_data.column(entry.name()).accept(visitor);
+        // HACK: Ignore missing columns.
+        auto &[success, column] = input_data.column_if_exists(entry.name());
+        if (success) {
+            column.accept(visitor);
```
alexdewar, line 339: If you follow my suggestion of using `std::optional`, this could become:

```suggestion
        // HACK: Ignore missing columns.
        if (auto column = input_data.column_if_exists(entry.name())) {
            column->accept(visitor);
```

src/HealthGPS/energy_balance_model.cpp
```diff
@@ -78,131 +82,169 @@ void EnergyBalanceModel::update_risk_factors(RuntimeContext &context) {
...
-    double RMR = 9.99 * BW_0 + 6.25 * H * 100.0 - 4.92 * person.age;
-    RMR += person.gender == core::Gender::male ? 5.0 : -161.0;
-    RMR *= 4.184; // kcal to kJ
+    // Compute resting metabolic rate (Mifflin-St Jeor).
+    double RMR = compute_RMR(BW_0, H, person.age, person.gender);
 
     double delta_0 = ((1.0 - beta_TEF) * PAL - 1.0) * RMR / BW_0;
 
-    // Update thermic effect of food and adaptive thermogenesis.
+    // Compute thermic effect of food and adaptive thermogenesis.
```
alexdewar, line 137: I think this can still be split up a bit further. I think this function should basically not do any computation directly and should just call a series of `compute_*` functions then bundle the results together.

alexdewar (changes requested):
Getting there! I've made some more small suggestions. I also think `simulate_person` could be split up a bit more.

src/HealthGPS/energy_balance_model.cpp
jamesturner246: I had this in mind, actually. I just want Ali to go over it and confirm all the units first, though. Some of them aren't so obvious.

src/HealthGPS/energy_balance_model.cpp
jamesturner246: It's quite a cheap op anyway. I'll just leave it for now.

src/HealthGPS/energy_balance_model.cpp
jamesturner246: Agree in principle. Thing I wonder is how to make `Person` hold a risk factor struct that's generic enough to handle all models, 1 static and 2 dynamic. I feel like we might fall back into the trap of having to save different state objects, with accompanying getter/setter/support code, for each different model, like in that `CachedRepository` issue. Would like to keep risk factor models as decoupled as possible, ideally.

Maybe the getter/setter/support code can all go in `EnergyBalanceModel`, though? How do you see it?

alexdewar: Ah yes, of course. I wasn't thinking straight.

You could add getters and setters to `EnergyBalanceModel` if you like. Could just be a freestanding function in `energy_balance_model.cpp` too. I just thought it would be cleaner to encapsulate this bit of boilerplate somehow.

alexdewar:
Btw I've re-requested @dalonsoa's review because he requested changes so it won't let you merge otherwise

src/HealthGPS/energy_balance_model.cpp
```diff
@@ -225,8 +221,8 @@ double EnergyBalanceModel::compute_EI(
     const std::unordered_map<core::Identifier, double> &nutrient_intakes) const {
     double EI = 0.0;
 
-    for (const auto &coefficient : energy_equation_) {
-        double delta_energy = nutrient_intakes.at(coefficient.first) * coefficient.second;
+    for (const auto &[nutrient_key, energy_coefficient] : energy_equation_) {
+        double delta_energy = nutrient_intakes.at(nutrient_key) * energy_coefficient;
         EI += delta_energy;
```
alexdewar:
```suggestion
        EI += nutrient_intakes.at(nutrient_key) * energy_coefficient;
```

src/HealthGPS.Core/datatable.cpp
```diff
@@ -46,15 +46,13 @@ const DataTableColumn &DataTable::column(const std::string &name) const {
     throw std::out_of_range(fmt::format("Column name: {} not found.", name));
 }
 
-const std::pair<bool, DataTableColumn &>
+const std::optional<std::reference_wrapper<const DataTableColumn>>
```
alexdewar, line 49: The `std::optional` bit doesn't need to be `const`. You don't want the user to muck about with the `DataTableColumn` inside, but they can reassign the returned value etc.

```suggestion
std::optional<std::reference_wrapper<const DataTableColumn>>
```

src/HealthGPS.Core/datatable.h
```diff
@@ -50,7 +50,8 @@ class DataTable {
     /// @brief Gets the column by name, also checking if the column exists
     /// @param name The column name
     /// @return A pair containing a success flag, and the column instance on success.
-    const std::pair<bool, DataTableColumn &> column_if_exists(const std::string &name) const;
+    const std::optional<std::reference_wrapper<const DataTableColumn>>
```
alexdewar, line 53: As above:

```suggestion
    std::optional<std::reference_wrapper<const DataTableColumn>>
```

alexdewar (changes requested):
I've made some v small suggestions, but otherwise I think we're good to go!

alexdewar (approved):
LGTM :smile:

PR #158, opened 2023-07-14
Use TBB's concurrent queue implementation

alexdewar:


jamesturner246 (approved):
Good. Lots of stuff deleted. So this is almost a drop-in replacement for the existing event queue?

alexdewar:
@jamesturner246 Yep. It will behave in exactly the same way.

src/HealthGPS.Console/CMakeLists.txt
```diff
@@ -1,9 +1,9 @@
 find_package(fmt CONFIG REQUIRED)
 find_package(cxxopts CONFIG REQUIRED)
 find_package(nlohmann_json CONFIG REQUIRED)
+find_package(TBB CONFIG REQUIRED)
 find_path(RAPIDCSV_INCLUDE_DIRS "rapidcsv.h")
 find_package(Threads REQUIRED)
```
dalonsoa, line 6: Knowing nothing about CMake, I'm curious: where does CMake looks for all of these packages?

dalonsoa (approved):
Looks good! Removing stuff that is not needed is always a good idea. It all seems to boil down to TBB. What is it about and why is no longer something that might not be found?

src/HealthGPS.Console/CMakeLists.txt
alexdewar: Traditionally, CMake looks for these on your system and e.g. the TBB library will come with a CMake package config file. And traditionally this worked rather poorly on Windows :stuck_out_tongue:.

Health-GPS uses [vcpkg](https://vcpkg.io/) on top to manage dependencies so they are automatically built from source before everything else is (provided you have vcpkg installed!).

alexdewar:
TBB is Intel's Threading Building Blocks library: https://github.com/oneapi-src/oneTBB

It basically just provides a bunch of useful bits for multithreaded applications, including tools for parallelising work and threadsafe containers (which I'm using here). Newer versions of the C++ standard include parallelised versions of common algorithms (e.g. sort) and gcc's implementation just uses TBB under the hood, so we need it for that anyway, so it kind of makes sense to just have it as a "proper" dependency and then we can use some of the nice features on all platforms.

Issue #89, opened 2022-11-04
[OSB] Suggestion: Use clang-tidy for static analysis


alexdewar:
I was hoping I'd have enough time to actually run it myself, but alas there were too many other things to be getting on with!

`clang-tidy` is an excellent static analysis tool, with a huge library of checks it can run: https://clang.llvm.org/extra/clang-tidy/

Some of the checks are for outright bugs, such as using an object after you've ``std::move``d from it, but it can also make suggestions about readability and performance. It is worth having a look through these and seeing which checks you think would be useful for your own code.

It integrates well with CMake: it can use the generated ``compile_commands.json`` file to find your source files, external headers etc.

In the past I have used it in conjunction with a CI system, but as a first step you could try just running it yourself over the code to see what issues it identifies. (You may need to fix any clang-related compile issues first.)

For reference, here is the CI script I used, including the tests that I used on my own code: https://github.com/BrainsOnBoard/bob_robotics/blob/master/bin/run_clang_tidy_check

PR #160, opened 2023-07-14
Fix: Throw error if invalid model specified (+ refactor)

alexdewar:
I've been playing around with `clang-tidy` as part of my work on #89 and it flagged a problem in `model_parser.cpp`. The issue is that if the JSON file contains an invalid model name, although an error message is printed, the code doesn't abort and will happily register a `nullptr` model definition in the cache.

I think part of the problem is that the `register_risk_factor_model_definitions` is rather long and with a lot of different branches in it, so it's easy to lose track of what the code is doing. I've split up the model loading into a series of separate functions with appropriate error handling in each.

### DIFF
--- a/src/HealthGPS.Console/model_parser.h
+++ b/src/HealthGPS.Console/model_parser.h
@@ -17,12 +17,14 @@ hgps::BaselineAdjustment load_baseline_adjustments(const poco::BaselineInfo &inf
 /// @brief Loads the full hierarchical linear regression model definition from a JSON file
 /// @param opt The parsed model definition JSON file
 /// @return An instance of the hgps::HierarchicalLinearModelDefinition type
+/// @throw std::invalid_argument if static model is unrecognised
 std::unique_ptr<hgps::RiskFactorModelDefinition>
 load_static_risk_model_definition(const poco::json &opt);
 
 /// @brief Loads a dynamic model from a JSON file
 /// @param opt The parsed model definition JSON file
-/// @return An instance of the hgps::LiteHierarchicalModelDefinition type
+/// @return An instance of the hgps::RiskFactorModelDefinition type
+/// @throw std::invalid_argument if dynamic model is unrecognised
 std::unique_ptr<hgps::RiskFactorModelDefinition>
 load_dynamic_risk_model_definition(const poco::json &opt);
 
@@ -51,6 +53,7 @@ load_risk_model_definition(const std::string &model_type, const poco::json &opt,
 /// @brief Load and parse the model file
 /// @param model_filename The path to the model
 /// @return The parsed JSON
+/// @throw std::invalid_argument if file is missing
 poco::json load_json(const std::string &model_filename);
 
 /// @brief Registers a risk factor model definition with the repository
--- a/src/HealthGPS.Console/model_parser.h
+++ b/src/HealthGPS.Console/model_parser.h
@@ -5,6 +5,8 @@
 #include "jsonparser.h"
 #include "options.h"
 
+#include <utility>
+
 namespace host {
 
 /// @brief Loads baseline adjustments information from a file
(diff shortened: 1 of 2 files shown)


dalonsoa (approved):
Looks good to me! Even if, for some reason, this does not solve all the problem, splitting long functions with a lot of nested branches is always a good idea.

src/HealthGPS.Console/model_parser.h
```diff
@@ -15,22 +17,42 @@ hgps::BaselineAdjustment load_baseline_adjustments(const poco::BaselineInfo &inf
 /// @brief Loads the full hierarchical linear regression model definition from a JSON file
 /// @param opt The parsed model definition JSON file
 /// @return An instance of the hgps::HierarchicalLinearModelDefinition type
-std::unique_ptr<hgps::HierarchicalLinearModelDefinition>
+std::unique_ptr<hgps::RiskFactorModelDefinition>
 load_static_risk_model_definition(const poco::json &opt);
```
jamesturner246, line 21:
```suggestion
load_static_risk_model_definition(const std::string &model_name, const poco::json &opt);
```

src/HealthGPS.Console/model_parser.h
```diff
@@ -15,22 +17,42 @@ hgps::BaselineAdjustment load_baseline_adjustments(const poco::BaselineInfo &inf
...
 
-/// @brief Loads the lite hierarchical linear regression model definition from a JSON file
+/// @brief Loads a dynamic model from a JSON file
 /// @param opt The parsed model definition JSON file
 /// @return An instance of the hgps::LiteHierarchicalModelDefinition type
-std::unique_ptr<hgps::LiteHierarchicalModelDefinition>
+std::unique_ptr<hgps::RiskFactorModelDefinition>
 load_dynamic_risk_model_definition(const poco::json &opt);
```
jamesturner246, line 27:
```suggestion
load_dynamic_risk_model_definition(const std::string &model_name, const poco::json &opt,
                                   const poco::SettingsInfo &settings);
```

src/HealthGPS.Console/model_parser.h
```diff
@@ -15,22 +17,42 @@ hgps::BaselineAdjustment load_baseline_adjustments(const poco::BaselineInfo &inf
...
+/// @param opt The parsed model definition JSON file
+/// @param settings The main model settings
+/// @return A std::pair containing the model type and model definition
+std::pair<hgps::HierarchicalModelType, std::unique_ptr<hgps::RiskFactorModelDefinition>>
+load_risk_model_definition(const std::string &model_type, const poco::json &opt,
+                           const poco::SettingsInfo &settings);
+
+/// @brief Load and parse the model file
+/// @param model_filename The path to the model
+/// @return The parsed JSON
```
jamesturner246, line 53:
```suggestion
/// @return The parsed JSON
/// @throw std::invalid_argument if file is missing
```

src/HealthGPS.Console/model_parser.h
```diff
@@ -15,22 +17,42 @@ hgps::BaselineAdjustment load_baseline_adjustments(const poco::BaselineInfo &inf
 /// @brief Loads the full hierarchical linear regression model definition from a JSON file
 /// @param opt The parsed model definition JSON file
 /// @return An instance of the hgps::HierarchicalLinearModelDefinition type
```
jamesturner246, line 19:
```suggestion
/// @return An instance of the hgps::HierarchicalLinearModelDefinition type
/// @throw std::invalid_argument if static model is unrecognised
```

src/HealthGPS.Console/model_parser.h
```diff
@@ -15,22 +17,42 @@ hgps::BaselineAdjustment load_baseline_adjustments(const poco::BaselineInfo &inf
...
-std::unique_ptr<hgps::HierarchicalLinearModelDefinition>
+std::unique_ptr<hgps::RiskFactorModelDefinition>
 load_static_risk_model_definition(const poco::json &opt);
 
-/// @brief Loads the lite hierarchical linear regression model definition from a JSON file
+/// @brief Loads a dynamic model from a JSON file
 /// @param opt The parsed model definition JSON file
 /// @return An instance of the hgps::LiteHierarchicalModelDefinition type
```
jamesturner246, line 25:
```suggestion
/// @return An instance of the hgps::RiskFactorModelDefinition type
/// @throw std::invalid_argument if dynamic model is unrecognised
```

jamesturner246 (changes requested):
Yeah, that looks much nicer. Just a couple of trivial header fixups and it's good to go!

src/HealthGPS.Console/model_parser.h
alexdewar: Oops! Good spot

alexdewar: Oops! Good spot

jamesturner246 (approved):
Looks good. Fire away!

PR #163, opened 2023-07-18
Add `clang-tidy` to config and add to CI

### LINKED ISSUE #89: [OSB] Suggestion: Use clang-tidy for static analysis

alexdewar:
This PR adds a basic `clang-tidy` config file with a minimal set of checks and uses a GitHub Action to use this to lint code in PRs. Note that this Action won't lint existing code -- just the bits introduced by the PR -- so we will still need to run `clang-tidy` locally when we add new checks. I've added the `run-clang-tidy.py` script to help with this; it just lets you run `clang-tidy` in parallel.

I've only enabled a small number of checks for now, because fixing up all the other issues `clang-tidy` can identify (e.g. `const`-ifying methods where possible) would cause a lot of code churn. Nonetheless, it did find a few small issues which I've fixed in this PR and #158. Once this is merged I'll incrementally add checks and fix up the code in separate PRs.

Closes #89.


run-clang-tidy.py
```diff
@@ -0,0 +1,503 @@
+#!/usr/bin/env python3
```
jamesturner246, line 1: On my machine, at least,there's already a `run-clang-tidy` shipped. I can't remember which package. What's the reason to include this here also?

.github/workflows/ci.yml
```diff
@@ -88,10 +88,30 @@ jobs:
           cd vcpkg
           ./bootstrap-vcpkg.sh -disableMetrics
 
+      - name: Configure HealthGPS (debug)
```
jamesturner246, line 91: Might ask to go over this with you in future. Starting to get quite big, and I'm not totally comfortable with the actions files yet.

run-clang-tidy.py
alexdewar: Ha, no reason. It turns out I have it too and didn't realise. I'll drop this file then

.github/workflows/ci.yml
alexdewar: Sure... We could chat tomorrow if you like?

jamesturner246: Sure, if you have a moment. Thanks

jamesturner246 (approved):
I'd say go ahead, once the Python script is dropped. Thanks!

alexdewar:
Thanks for review!

PR #168, opened 2023-07-18
Add food group unit price and nutrient range config, load into EBM model

### LINKED ISSUE #132: Add nutrient boundaries to new energy balance model config.json
Nutrient values in the new energy balance model need to be bounded to within a given range.

This would be implemented by adding the boundaries to the `Nutrients` field in e.g. `./example/France.NewEBM.json`, like this:

```json
        ...
        "Nutrients": {
                "Carbs": [0.5, 0.75,
                "Fat": [0.3, 0.6],
                "Protein: [0.6, 0.6]",
                "Sodium: [0.01, 0.1]"
        },
        ...
```

These boundaries would then be loaded in model_parser.cpp into the new `EnergyBalanceModel` class, for use in simulation. Code would need to be added to enforce these boundaries after total nutrients are calculated.

### LINKED ISSUE #136: Add `"Price"` field to `"Foods"` array entries
In new EBM config file `./example/France.NewEBM.json`, add `"Price"` field for each `"Foods"` entry.

```json
...
            {
                    "Name": "Sandwich",
                    "Price": 4.00,
                    "Nutrients": {
                            "Fat": 1.0,
                            ...
                    }
            },
 ...
```

This value is constant (requires data). Just use dummy values for now.

Needs code in `model_parser.cpp` to load these values from JSON, and code in `EnergyBalanceModelDefinition` to load it into the model.

jamesturner246:
This one should be straightforward. I've added nutrient `"Range"`s and food group unit `"Price"`s to the new EBM `France.NewEBM.json`, added `model_parser` and `EnergyBalanceModel` to load and sanity-check the values, and added a method to `EnergyBalanceModel` to return a range-bounded nutrient value.

Fixes #136 
Fixes #132


example/France.NewEBM.json
```diff
@@ -9,42 +9,50 @@
...
         {
             "Name": "Sodium",
+            "Range": [0.0, 50.0],
             "Energy": 0.0
         }
     ],
     "Foods": [
         {
             "Name": "Carbs",
+            "Price": 0.0,
```
alexdewar, line 34: Should this maybe be `null` instead? Carbs ain't free, ya know :stuck_out_tongue:

src/HealthGPS/energy_balance_model.cpp
```diff
@@ -81,6 +88,18 @@ void EnergyBalanceModel::update_risk_factors(RuntimeContext &context) {
...
+double EnergyBalanceModel::bounded_nutrient_value(const core::Identifier &nutrient,
+                                                  double value) const {
+    auto &range = nutrient_ranges_.at(nutrient);
+    if (value < range.first) {
+        return range.first;
+    }
+    if (value > range.second) {
+        return range.second;
+    }
+    return value;
```
alexdewar, line 100: Are you sure you want to cap the values rather than throwing an error if the values are out of bounds?

If so, you can write this as so:

```suggestion
    const auto &range = nutrient_ranges_.at(nutrient)
    return std::clamp(range.first, range.second, value);
```

You need to `#include <algorithm>` for this.

jamesturner246: This is an interesting question, and one that should be bought up next meeting. The existing code merely caps without error. Maybe there's some argument of homeostasis or digestion capacity. I'll bring it up.

alexdewar: Maybe open a GitHub issue for it too so we don't forget about it?

From the bits of the code I've looked at, it seems that doing this sort of thing (i.e. carrying on in the face of unexpected input) is fairly common, but I don't think it's good. It'd be better to give the user a big fat error message and then stop IMO, or at least give them a warning!

src/HealthGPS.Console/jsonparser.h
```diff
@@ -81,4 +82,25 @@ void from_json(const json &j, PolicyScenarioInfo &p);
 // Output information
 void to_json(json &j, const OutputInfo &p);
 void from_json(const json &j, OutputInfo &p);
+
 } // namespace host::poco
+
+namespace std {
+
+// Optional parameters
+template <typename T> void to_json(nlohmann::json &j, const std::optional<T> &p) {
```
alexdewar, line 91: Good templating!

I'm a little surprised the library doesn't include this by default...

alexdewar (approved):
Looking good. I think we're ready to merge :smile:.

PR #175, opened 2023-07-20
Only run clang-tidy action on PRs

### LINKED ISSUE #174: `clang-tidy` GitHub Action causing CI failures on `main`
Example: https://github.com/imperialCHEPI/healthgps/actions/runs/5608727052/jobs/10261468237

The problem seems to be that the action needs to be triggered by a PR, so if it's just triggered by a push to `main` instead, then it fails with an error, which is not ideal.

alexdewar:
Otherwise it causes spurious failures.

Fixes #174.

It's a pity that clang-tidy isn't also run on pushes to main, but probably not the end of the world
as things will only be submitted via PRs anyway.


jamesturner246 (approved):
Okay, makes sense.

We *could* lock main, such that only code from PRs can merge into main. I think I saw the option whilst I was configuring branch protection.

alexdewar:
I just checked and that already seemed to be enabled. Maybe I enabled it myself? Dunno. Either way, I think it's sensible!

PR #176, opened 2023-07-20
Add docs source to main branch

### LINKED ISSUE #71: [OSB] Suggestion: Technical documentation should be in main branch
It makes sense logically to have these changes in the same place as the code; often if you change some aspect of the code you will also need to make a change to the documentation and it makes sense to do this in a single commit.

You could get GitHub to use your ``/docs`` folder for documentation to start with. This will require a bit more tweaking if/when you implement #69 and want to include documentation generated by e.g. Doxygen.

### LINKED ISSUE #165: Move most developer instructions from README to developer guide
I think it's fine to have basic build instructions in there, but probably the stuff about `pre-commit` should be along with the rest of the developer documentation. See: https://imperialchepi.github.io/healthgps/development

It probably makes sense to do #71 first.

alexdewar:
This PR moves the docs over from the separate `gh-pages` branch into `main` and makes the necessary changes to the CI to build both the Jekyll (markdown) documentation and API documentation with doxygen. I'm using the new GitHub Actions workflow for doing this: https://github.blog/changelog/2022-07-27-github-pages-custom-github-actions-workflows-beta/

I haven't changed the content much, though I did:

- Fix up places where the documentation was now out of date (this is why your docs should live with your code...)
- Fix up all the `markdownlint` warnings
- Move developer-specific instructions from the README to the documentation proper (#165).

One other change is that now the doxygen documentation will be regenerated with every push to `main`, whereas before it was only updated with new releases. I think this is the right way to do things though. We could consider making a PDF of the API documentation as a release asset too, but I'm not sure it's needed.

Fixes #71. Fixes #165.


jamesturner246 (commented):
I'm a bit concerned about regenerating documentation on PR merge to main when HealthGPS has point releases, unless we are able to somehow publish 'development' and 'release' versions of the documentation, and default to 'release'.

Might have problems down the line when the doc pages describe 'my fancy new feature', but it isn't available in the most recent point release. Unless we make a point of (and remember to) only update the docs in the final PR before a release. Any other options?

alexdewar:
That's a good point. We could set it to just build the documentation on a release event, rather than on every push to `main`. The downside is that then it would be difficult to do documentation hotfixes, though I suppose there are ways to get round this.

Would you like me to do this?

jamesturner246:
I think it's maybe the lesser of the two evils. Yes please. Maybe later, ultimately, we could build develop and release documentations, but in the meantime...

jamesturner246 (approved):
Thanks Alex. Go ahead.

PR #182, opened 2023-07-20
Enable `bugprone-*` checks for `clang-tidy`

### LINKED ISSUE #177: `clang-tidy`: Add `bugprone-*` checks

alexdewar:
This PR enables and fixes most of `clang-tidy`'s `bugprone-*` checks, which warn about use of bugprone constructs in code.

I decided not to enable `bugprone-narrowing-conversions` for now, as it was giving a lot of warnings about converting ages between signed and unsigned `int`s. The warnings are sensible and we should be consistent (though it's not likely to be a problem unless someone's age is more than 2^31) but I think it's a job for another day.

The main warnings it complained about were:

1. Accessing `std::optional`s' values without checking whether they had one
2. Functions that were erroneously marked as `noexcept`

Fixes #177.


jamesturner246 (approved):
Looks okay -- tests are passing. Go ahead!

I've added an issue for the narrowing conversions.

Issue #161, opened 2023-07-14
Sanitisation of input data


jamesturner246:
The whole model loading core in `model_parser` could do with a rethink, to test that both the main config file -- e.g. `France.Config.json` -- has sensible values and if fully defined. Same for the static and dynamic risk factor configs -- e.g. `France.HLM.json` and `France.NewEBM.json`.

Furthermore, we should check that the data is complete over defined ranges. For example, if max age is defined as x, we should ensure that all age dependent data is defined up to at least x.

The code should fail on any missing data, and any missing files -- it often just carries on silently/with a warning message.

PR #189, opened 2023-07-24
Remove dynamic risk factor mechanism

### LINKED ISSUE #183: Remove unused 'dynamic risk factor' mechanism
After discussions with Israel, it seems that the 'dynamic risk factor' mechanism is no longer used, and can be removed.

It was used in a prior regression model which relied on the simulated year, but this is no longer needed.

jamesturner246:
Should be straightforward. Just a bunch of deleted code for the unused dynamic risk factor mechanism that is no longer needed in the hierarchical linear regression models.

I'll include @israel-vieira in the loop, if interested. Obviously no commitment to a review expected though!

Fixes #183

### DIFF
--- a/src/HealthGPS.Console/configuration.cpp
+++ b/src/HealthGPS.Console/configuration.cpp
@@ -210,24 +210,6 @@ Configuration load_configuration(CommandOptions &options) {
             }
         }
 
-        if (!config.modelling.dynamic_risk_factor.empty()) {
-            auto found_dynamic_factor = false;
-            for (auto &factor : config.modelling.risk_factors) {
-                if (core::case_insensitive::equals(factor.name,
-                                                   config.modelling.dynamic_risk_factor)) {
-                    found_dynamic_factor = true;
-                    break;
-                }
-            }
-
-            if (!found_dynamic_factor) {
-                fmt::print(
-                    fg(fmt::color::red),
-                    "\nInvalid configuration, dynamic risk factor: {} is not a risk factor.\n",
-                    config.modelling.dynamic_risk_factor);
-            }
-        }
-
         // Run-time
         opt["running"]["start_time"].get_to(config.start_time);
         opt["running"]["stop_time"].get_to(config.stop_time);
@@ -353,21 +335,12 @@ ModelInput create_model_input(core::DataTable &input_table, core::Country countr
 
     auto mapping = std::vector<MappingEntry>();
     for (auto &item : config.modelling.risk_factors) {
-        if (core::case_insensitive::equals(item.name, config.modelling.dynamic_risk_factor)) {
-            if (item.range.empty()) {
-                mapping.emplace_back(
-                    MappingEntry(item.name, item.level, core::Identifier{item.proxy}, true));
-            } else {
-                auto boundary = FactorRange{item.range[0], item.range[1]};
-                mapping.emplace_back(MappingEntry(item.name, item.level,
-                                                  core::Identifier{item.proxy}, boundary, true));
-            }
-        } else if (!item.range.empty()) {
+        if (item.range.empty()) {
+            mapping.emplace_back(MappingEntry{item.name, item.level, core::Identifier{item.proxy}});
+        } else {
             auto boundary = FactorRange{item.range[0], item.range[1]};
             mapping.emplace_back(
                 MappingEntry{item.name, item.level, core::Identifier{item.proxy}, boundary});
-        } else {
-            mapping.emplace_back(MappingEntry{item.name, item.level, core::Identifier{item.proxy}});
         }
     }
 
--- a/src/HealthGPS.Console/configuration.cpp
+++ b/src/HealthGPS.Console/configuration.cpp
@@ -336,11 +336,10 @@ ModelInput create_model_input(core::DataTable &input_table, core::Country countr
     auto mapping = std::vector<MappingEntry>();
     for (auto &item : config.modelling.risk_factors) {
         if (item.range.empty()) {
-            mapping.emplace_back(MappingEntry{item.name, item.level, core::Identifier{item.proxy}});
+            mapping.emplace_back(item.name, item.level, item.proxy);
         } else {
             auto boundary = FactorRange{item.range[0], item.range[1]};
-            mapping.emplace_back(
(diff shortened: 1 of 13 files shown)


dalonsoa (approved):
Is this model something truly obsolete that will not be needed anymore? Or is it still potentially useful but does not fit within the new way of defining models?

I'm just asking because deleting unused things is always good... as long as it is certain they will no longer be needed.

jamesturner246:
I discussed it with Israel, also included here, and the consensus seemed to be exactly that. I'm to understand that neither models use it, nor will use it.

src/HealthGPS.Console/configuration.cpp
```diff
@@ -353,21 +335,12 @@ ModelInput create_model_input(core::DataTable &input_table, core::Country countr
...
-                mapping.emplace_back(MappingEntry(item.name, item.level,
-                                                  core::Identifier{item.proxy}, boundary, true));
-            }
-        } else if (!item.range.empty()) {
+        if (item.range.empty()) {
+            mapping.emplace_back(MappingEntry{item.name, item.level, core::Identifier{item.proxy}});
+        } else {
             auto boundary = FactorRange{item.range[0], item.range[1]};
             mapping.emplace_back(
                 MappingEntry{item.name, item.level, core::Identifier{item.proxy}, boundary});
```
alexdewar, line 343: I think this could be just:

```suggestion
        if (item.range.empty()) {
            mapping.emplace_back(item.name, item.level, item.proxy);
        } else {
            auto boundary = FactorRange{item.range[0], item.range[1]};
            mapping.emplace_back(item.name, item.level, item.proxy, boundary);
```

src/HealthGPS.Tests/HealthGPS.Test.cpp
```diff
@@ -17,7 +17,7 @@ namespace fs = std::filesystem;
 
 static std::vector<hgps::MappingEntry> create_mapping_entries() {
     using namespace hgps;
-    return std::vector<MappingEntry>{MappingEntry("Year", 0, core::Identifier::empty(), true),
+    return std::vector<MappingEntry>{MappingEntry("Year", 0, core::Identifier::empty()),
```
alexdewar, line 20: I think the compiler can infer the type from the return type here:

```suggestion
    return {MappingEntry("Year", 0, core::Identifier::empty()),
```

(You'll need to `clang-format` this.)

src/HealthGPS.Tests/HierarchicalMapping.Test.cpp
```diff
@@ -4,7 +4,7 @@
 
 static std::vector<hgps::MappingEntry> create_mapping_entries() {
     using namespace hgps;
-    return std::vector<MappingEntry>{MappingEntry("Year", 0, core::Identifier::empty(), true),
+    return std::vector<MappingEntry>{MappingEntry("Year", 0, core::Identifier::empty()),
```
alexdewar, line 7: Ditto above.

src/HealthGPS/mapping.cpp
```diff
@@ -7,24 +7,17 @@
...
-hgps::MappingEntry::MappingEntry(std::string name, short level, core::Identifier entity_key,
-                                 FactorRange range, bool dynamic_factor)
-    : name_{name}, name_key_{core::Identifier{name}}, level_{level},
-      entity_key_{std::move(entity_key)}, range_{range}, dynamic_factor_{dynamic_factor} {}
-
-MappingEntry::MappingEntry(std::string name, short level, core::Identifier entity_key,
-                           bool dynamic_factor)
-    : MappingEntry(name, level, entity_key, FactorRange{}, dynamic_factor) {}
 
 MappingEntry::MappingEntry(std::string name, short level, core::Identifier entity_key,
```
alexdewar, line 11: What on earth are we using `short`s for?! :laughing:

alexdewar (approved):
A few small suggestions, but otherwise LGTM. Merge whenever you're done.

Are you finished with refactoring the config code for the time being btw? Is now a good time for me to work on #161?

src/HealthGPS/mapping.cpp
jamesturner246: Memory efficiency! xD I have no idea.

jamesturner246:
I will be removing e.g, the risk factors from the main config file, but it's up to you. If so, I'll try to be careful and synchronise.

src/HealthGPS.Tests/HealthGPS.Test.cpp
israel-vieira, line 1: Remove the "Year" variable from the list to remove the time-dynamic dimension from the tests.

src/HealthGPS.Tests/HierarchicalMapping.Test.cpp
israel-vieira, line 1: Remove the "Year" variable from the list to remove the time-dynamic dimension from the tests.

src/HealthGPS.Tests/HealthGPS.Test.cpp
israel-vieira, line 1: Remove the "Year" variable from the list to remove the time-dynamic dimension from the tests.

israel-vieira (approved):
**Looks good**, just two minor suggestions.

**Note:**

The dynamic factor was originally defined as part of a generic regression framework to represent dynamic risk factor for adults, it included a _time dimension_, usually year in the regression variables to move the population over time. This model was dropped in favor of the current EB model for children, never used, and the code removed from the main branch.

The new EB model is not regression-based, it is safe to remove the dynamic risk factor concept from the source code and configuration.

alexdewar:
@jamesturner246 Ok cool. For future reference, I'd probably have updated the config files in this PR too just to make sure the code still works. It's fine to do it later though. Maybe open an issue?

I'm happy to hold off on #161 until you've made those changes.

jamesturner246:
@alexdewar , the modifications to the config file required for this change are already done. I simply removed `dynamic_risk_factor`. The remaining changes of the config are for other PRs.

src/HealthGPS.Tests/HierarchicalMapping.Test.cpp
jamesturner246: Thanks @israel-vieira. I've removed year from both places.

PR #190, opened 2023-07-24
Don't build docs by default

alexdewar:
Currently `BUILD_DOC` is set to `ON` by default, though curiously the option only works if you're building in release mode. As most users probably don't want to build it locally, let's disable it by default and make it an opt-in feature. I've added a description of how to build the documentation locally for users who want to do so.


jamesturner246 (approved):
Straightforward and documented. Go ahead.

PR #192, opened 2023-07-29
Remove 'proxy' risk factorm key mechanism 

### LINKED ISSUE #184: Remove 'proxy' variable name mechanism, if possible
This mechanism is used in the 'dynamic risk factor' mechanism, which is no longer used (see #183).

I *believe* this is no longer used now, but investigate and remove if possible.

jamesturner246:
Turns out we can remove the 'proxy' risk factor name mechanism now without worry. It only mattered when your risk factor was named differently from how it was named in our `Person` structs, for instance.

They were identically named in the example code. In terms of future usefulness, I really can't see any use case of naming your risk factors and corresponding `Person` attributes differently. Nonetheless, I will await the reviewers' opinions!

Fixes #184

### DIFF
--- a/example/France.Config.json
+++ b/example/France.Config.json
@@ -36,52 +36,56 @@
             {
                 "name": "Gender",
                 "level": 0,
-                "proxy": "gender",
                 "range": [0, 1]
             },
-            { "name": "Age", "level": 0, "proxy": "age", "range": [1, 87] },
-            { "name": "Age2", "level": 0, "proxy": "", "range": [1, 7569] },
-            { "name": "Age3", "level": 0, "proxy": "", "range": [1, 658503] },
+            {
+                "name": "Age",
+                "level": 0,
+                "range": [1, 87]
+            },
+            {
+                "name": "Age2",
+                "level": 0,
+                "range": [1, 7569]
+            },
+            {
+                "name": "Age3",
+                "level": 0,
+                "range": [1, 658503]
+            },
             {
                 "name": "SES",
                 "level": 0,
-                "proxy": "ses",
                 "range": [-2.316299, 2.296689]
             },
             {
                 "name": "Sodium",
                 "level": 1,
-                "proxy": "",
                 "range": [1.127384, 8.656519]
             },
             {
                 "name": "Protein",
                 "level": 1,
-                "proxy": "",
                 "range": [43.50682, 238.4145]
             },
             {
                 "name": "Fat",
                 "level": 1,
-                "proxy": "",
                 "range": [45.04756, 382.664098658922]
             },
             {
                 "name": "PA",
                 "level": 2,
-                "proxy": "",
                 "range": [22.22314, 9765.512]
             },
             {
                 "name": "Energy",
                 "level": 2,
-                "proxy": "",
                 "range": [1326.14051277588, 7522.496]
             },
             {
                 "name": "BMI",
                 "level": 3,
-                "proxy": "",
                 "range": [13.88, 39.48983]
             }
         ],
--- a/src/HealthGPS/mapping.cpp
+++ b/src/HealthGPS/mapping.cpp
@@ -9,7 +9,7 @@
 namespace hgps {
 
 MappingEntry::MappingEntry(std::string name, int level, FactorRange range)
-    : name_{name}, name_key_{core::Identifier{name}}, level_{level}, range_{range} {}
+    : name_{std::move(name)}, name_key_{name_}, level_{level}, range_{range} {}
 
 const std::string &MappingEntry::name() const noexcept { return name_; }
 
--- a/src/HealthGPS/mapping.cpp
+++ b/src/HealthGPS/mapping.cpp
@@ -8,26 +8,12 @@
 
 namespace hgps {
 
-MappingEntry::MappingEntry(std::string name, int level, core::Identifier entity_key,
-                           FactorRange range)
-    : name_{name}, name_key_{core::Identifier{name}}, level_{level},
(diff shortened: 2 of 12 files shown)


src/HealthGPS/mapping.cpp
```diff
@@ -8,27 +8,13 @@
...
-    : name_{name}, name_key_{core::Identifier{name}}, level_{level},
-      entity_key_{std::move(entity_key)}, range_{range} {}
-
-MappingEntry::MappingEntry(std::string name, int level, core::Identifier entity_key)
-    : MappingEntry(name, level, entity_key, FactorRange{}) {}
-
-MappingEntry::MappingEntry(std::string name, int level)
-    : MappingEntry(name, level, core::Identifier::empty(), FactorRange{}) {}
+MappingEntry::MappingEntry(std::string name, int level, FactorRange range)
+    : name_{name}, name_key_{core::Identifier{name}}, level_{level}, range_{range} {}
```
alexdewar, line 12: You're copying `name` twice, but you could do it with one move and one copy instead:

```suggestion
    : name_{std::move(name)}, name_key_{name_}, level_{level}, range_{range} {}
```

It slightly pains me that there are effectively two string identifiers in this class (`name_` and `name_key_`). I'm guessing the difference between them is effectively that the `core::Identifier` is always lowercase?

src/HealthGPS/mapping.h
```diff
@@ -51,20 +51,8 @@ class MappingEntry {
...
-    /// @param name Risk factor name
-    /// @param level The hierarchical level
-    /// @param entity_key The associated Person property, if exists
-    MappingEntry(std::string name, int level, core::Identifier entity_key);
-
-    /// @brief Initialises a new instance of the MappingEntry class
-    /// @param name Risk factor name
-    /// @param level The hierarchical level
-    MappingEntry(std::string name, int level);
+    MappingEntry(std::string name, int level, FactorRange range = {});
```
alexdewar, line 55: I'm guessing `FactorRange` defaults to `(0, 0)`. Is that safe or would it be better to use a `std::optional` for it, assuming that `(0, 0)` just indicates that there is no range specified?

This is possibly out of scope for this PR.

alexdewar (approved):
This looks sensible to me.

I've made a few minor suggestions -- all pretty superficial stuff.

example/France.Config.json
```diff
@@ -36,52 +36,56 @@
...
-            { "name": "Age", "level": 0, "proxy": "age", "range": [1, 87] },
-            { "name": "Age2", "level": 0, "proxy": "", "range": [1, 7569] },
-            { "name": "Age3", "level": 0, "proxy": "", "range": [1, 658503] },
+            {
+                "name": "Age",
+                "level": 0,
+                "range": [1, 87]
+            },
+            {
+                "name": "Age2",
```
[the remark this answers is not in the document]
jamesturner246: Age is age, age2 is precomputed age squared, age3 is precomputed age cubed. Even though I think the square and cube of age are still computed in code now.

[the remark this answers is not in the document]
alexdewar: Oh, well in that case shall we drop them then? Doesn't have to be in this PR.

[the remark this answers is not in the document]
jamesturner246: I'll do it in next PR, where I'm already messing with config and input files.

src/HealthGPS/mapping.h
jamesturner246: It defaults to the 'empty' range, which is practically identical to an optional. I might as well just swap it out for optional now.

alexdewar: Sounds good. It's nicer to have the extra type safety you get with `std::optional`.

israel-vieira:
> Turns out we can remove the 'proxy' risk factor name mechanism now without worry. It only mattered when your risk factor was named differently from how it was named in our `Person` structs, for instance.
> 
> _**They were identically named in the example code**_. In terms of future usefulness, I really can't see any use case of naming your risk factors and corresponding `Person` attributes differently. Nonetheless, I will await the reviewers' opinions!
> 
> Fixes #184

Hi James, sorry for being late for the review. 

There are many parts of the existing model that is relevant only to hierarchical regression models, however please use the _**example files**_ just as is, not as the only possible configuration. The **_proxy_** concept was used to link external fitted models, based on unknown survey datasets, no-standard variable names, etc., and the model's internal convention, common occurrences are: _sex_ vs. _gender_ and many flavors of age naming.

The _variables range_ is used to constrain the regression models from overshooting outside of the input dataset, not sure that the new EB model requires such mechanism. The age2 and age3 are also regression specific, their are included in the initial dataset, t0, but must be dynamically calculated to account for people aging over time.

Anyway, please double check with Ali about the new EBM requirements.

Kind regards,

Israel

jamesturner246:
Thanks for the info @israel-vieira. I can confirm that the modelling team still want the ability to bound risk factor variables to within ranges, and that functionality remains in the code. I've suggested an additional clipping method, using the logistic function, but the hard clipping method will remain.

I've been very careful to keep the original example model as is, and have a temporary new dynamic model file `France.NewEBM.json`, which I'm adding to as required. This will eventually be fleshed out into a separate example project, distinct from the existing model, which will remain in the example dir..

Regarding the variable naming conventions, I'm inclined to say that this would more appropriately be corrected in the pre-processing or model loading phase, rather than being exposed to the models via the mapping class. I'm happy to reintroduce the aliases for use during model loading/JSON parsing, but from here on I'm trying to make the code which the models themselves interact with as lightweight and simple as possible

Best wishes,
James.

Issue #194, opened 2023-08-02
Merge Static and Dynamic risk factors


jamesturner246:
I've put some thought into this, and it has become clear that the main thing hindering further refactoring -- thus simplifying the process of adding new RF models -- is the separation of Static and Dynamic risk factor models.  Some reasons:

* Both of the existing older RF model S/D pair use common functionality -- e.g. the `HierarchicalMapping` mechanism that's passed around everywhere, that is not necessarily useful for other RF implementations. This can instead be contained in the new merged RF model class.
* It is preventing separation of JSON risk factor config from the main config file, since, for example, both static and dynamic models rely on the same list of risk factors in the main config, and moving it from the main config would require duplicating them in the static/dynamic RF configs.
* Cleaner and more encapsulated code: a *lot of code for handling the separate S and D risk models can be removed, and instead placed inside the class which requires it.

I'll add to this list as I think of more, but it's already starting to become an attractive and clean solution in my head.

israel-vieira:
Do you have a new, non-regression based static model for initialization? Perhaps you could use a base class to minimize the duplication, but from memory, the existing static and dynamic have very different sampling methods.

jamesturner246:
From what I interpret from what Ali has said, there will be a new initialisation method, but you have a point that it's worth considering if this new, as yet unclear, method can be fit into the existing static HLM framework. I will try to get an answer to this in tomorrow's meeting.

Otherwise, despite sampling methods, from what, I can see, the core functionality is still separated into the `generate_risk_factors` and `update_risk_factors` methods. In the static model, `update_risk_factors` is implemented such that it only fires for `Person`s of age zero, and in that case it behaves identically to `generate_risk_factors`. In the dynamic model, only `update_risk_factors` is implemented, with `generate_risk_factors` left undefined.

israel-vieira:
In that case, I would wait until the new model structure is defined, try to understand the new model before doing any major consolidation work, that might not even be relevant to the new models.

The current two models MUST share the same risk factors because the first one, S, initialize the virtual population risk factors absolute values and the second, D, updates the population over time by applying deltas to the same risk factors.

The population size is dynamic, new born babies must be initialized by the S model at birth at any point in time during the simulation. The two models looks similar because of the regressions approach, but the fitting process, coefficients, input data, etc. are very different, just look at the size of the input files.

Anyway, focus on the new model, if the new approach is not regression, the existing code might be irrelevant,

PR #195, opened 2023-08-02
Risk factor refactor

### LINKED ISSUE #166: Decouple risk factor code from DataTable backend
Right now, `DataTable` initialisation (from CSV files) is very tightly coupled to the risk factors defined in the config JSON files. Adding risk factors tends to break the `DataTable` if corresponding CSV data columns have not been provided.

We should investigate the parts of the code that prevent this from adding non-data-driven risk factors which are not initialised with data from the DataTable (CSV files), and change it to allow us to choose how risk factors are individually initialised.

jamesturner246:
Fixes #166 
(not really)

This has been somewhat of a lacklustre attempt that fell short of my ambition... All that was really achieved is the removal of duplicate risk factor listings from `France.EBHLM.json`, and changing a couple of data manager methods to fail on missing data, rather than warn and continue. As such, it should be quite quick to review. The main issue still seems to be the presence of code for handling Hierarchical static and dynamic models in the wider codebase, which is resisting refactor efforts.

In a moment of catharsis, I've now convinced myself that the best solution going forward is to merge static and dynamic RF model into a single model, which encapsulates all this extra HLM-specific code within it. 

See #194. I'd appreciate your input, as I'm itching to get started merging them, rather than doing more piecemeal refactors that aren't solving the underlying problem.

### DIFF
--- a/example/France.EBHLM.json
+++ b/example/France.EBHLM.json
@@ -7,63 +7,6 @@
         "Alpha3": "FRA"
     },
     "BoundaryPercentage": 0.05,
-    "RiskFactors": [
-        {
-            "Name": "Gender",
-            "Level": 0,
-            "Range": [0, 1]
-        },
-        {
-            "Name": "Age",
-            "Level": 0,
-            "Range": [1, 87]
-        },
-        {
-            "Name": "Age2",
-            "Level": 0,
-            "Range": [1, 7569]
-        },
-        {
-            "Name": "Age3",
-            "Level": 0,
-            "Range": [1, 658503]
-        },
-        {
-            "Name": "SES",
-            "Level": 0,
-            "Range": [-2.316299, 2.296689]
-        },
-        {
-            "Name": "Sodium",
-            "Level": 1,
-            "Range": [1.127384, 8.656519]
-        },
-        {
-            "Name": "Protein",
-            "Level": 1,
-            "Range": [43.50682, 238.4145]
-        },
-        {
-            "Name": "Fat",
-            "Level": 1,
-            "Range": [45.04756, 382.664098658922]
-        },
-        {
-            "Name": "PA",
-            "Level": 2,
-            "Range": [22.22314, 9765.512]
-        },
-        {
-            "Name": "Energy",
-            "Level": 2,
-            "Range": [1326.14051277588, 7522.496]
-        },
-        {
-            "Name": "BMI",
-            "Level": 3,
-            "Range": [13.88, 39.48983]
-        }
-    ],
     "Variables": [
         {
             "Name": "dPA",
--- a/src/HealthGPS.Console/configuration.cpp
+++ b/src/HealthGPS.Console/configuration.cpp
@@ -293,7 +293,7 @@ std::vector<core::DiseaseInfo> get_diseases_info(core::Datastore &data_api, Conf
                config.diseases.size());
 
     for (const auto &code : config.diseases) {
-        result.emplace_back(data_api.get_disease_info(core::Identifier{code}));
+        result.emplace_back(data_api.get_disease_info(code));
     }
 
     return result;
--- a/src/HealthGPS.Console/configuration.cpp
+++ b/src/HealthGPS.Console/configuration.cpp
@@ -293,13 +293,7 @@ std::vector<core::DiseaseInfo> get_diseases_info(core::Datastore &data_api, Conf
                config.diseases.size());
 
     for (const auto &code : config.diseases) {
-        auto code_key = core::Identifier{code};
-        auto item = data_api.get_disease_info(code_key);
-        if (item.has_value()) {
-            result.emplace_back(item.value());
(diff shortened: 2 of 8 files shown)


src/HealthGPS.Console/configuration.cpp
israel-vieira, line 1: The list of diseases can be long, 30+. The current validation will list all the invalid diseases key, if any, before exiting. The exception alternative will break at the first invalid key, the use will need to keep on trying until all are correct, in the end, both methods are accomplishing the required validation,  it is just a matter of style.

Mapping the config with the back-end in a case-sensitive world is never user friend, the identifier type tries to minimize the pain by making all keys lowercase.

israel-vieira (approved):

example/France.EBHLM.json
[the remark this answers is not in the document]
jamesturner246: Agreed, the respective healthgps-tools script should also be updated. I'm considering a Python rewrite of the scripts, to sidestep the Windows R binding dependency.

src/HealthGPS.Console/configuration.cpp
jamesturner246: Case-sensitivity: fair point. In which case, the task then becomes removing the parts of the codebase that *are* case-sensitive. I've spotted a few.

Diseases: agree, a style choice. My concern in these cases has always been that by warning and continuing for dodgy input, that it makes the problem slightly less obvious to diagnose when it fails later on in the program, than if it were to immediately fail on loading. Perhaps this can be replaced with a function which reads all diseases, detects missing ones with printed debug info, before throwing an exception for *any* missing inputs, as opposed to a case-by-case exception.

dalonsoa: I'm with @israel-vieira on this - collecting all that is invalid and providing a full list to the user (within meaningful indication of what is wrong) results on a better user experience than having to try again and again until all issues are sorted. We have another project where this is the approach and is indeed much better. 

All in all, having a strong validation and input pre-processing steps where things can fail early and inconsistencies fixed (casing, proxis, etc.) is a good thing that will result in simpler, easier to understand (and diagnose) code.

israel-vieira: Personally, I don't consider users input validation as an exceptional circumstance, the validation should focus on helping the users to fix the wrong or missing information provided, perhaps the system should show valid options before exiting gracefully.

In general, users don't read or understand stack traces with a message hidden inside a large block of text, this is kind of information is more relevant for developers.

alexdewar (changes requested):
I think this is the right approach on the whole. I've made some minor suggestions.

I agree with @dalonsoa and @israel-vieira that we do probably want to do "lazy failure" when loading the config and print errors as we go along, rather than aborting as soon as an error is found. I think what you've done is an improvement though, so I'd be inclined to merge this as is for now and worry about that later. As I was planning on tackling this myself (see #161), I'm happy for you to leave this task for me, if you like.

src/HealthGPS.Tests/Datastore.Test.cpp
```diff
@@ -11,10 +11,7 @@ class DatastoreTest : public ::testing::Test {
     DatastoreTest() : manager{test_datastore_path} {}
 
     hgps::data::DataManager manager;
-
-    // We don't need to check because this is just for testing
-    hgps::core::Country uk =
-        manager.get_country("GB").value(); // NOLINT(bugprone-unchecked-optional-access)
+    hgps::core::Country uk = manager.get_country("GB");
```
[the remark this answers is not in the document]
jamesturner246: Same.

src/HealthGPS.Tests/Datastore.Test.cpp
```diff
@@ -130,15 +123,23 @@ TEST_F(DatastoreTest, CountryMortality) {
...
+
+TEST_F(DatastoreTest, GetDiseaseInfoMatchesGetDisases) {
     auto diseases = manager.get_diseases();
     for (auto &item : diseases) {
+        EXPECT_NO_THROW(manager.get_disease_info(item.code));
         auto info = manager.get_disease_info(item.code);
-        EXPECT_TRUE(info.has_value());
-        EXPECT_EQ(item.code, info.value().code); // NOLINT(bugprone-unchecked-optional-access)
+        EXPECT_EQ(item.code, info.code);
     }
```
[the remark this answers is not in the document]
jamesturner246: Good spot.

src/HealthGPS.Console/configuration.cpp
jamesturner246: Agree with this ^. The user gets helpful human readable messages noting which keys were missing, and execution stops before there's a chance of accidental continuation on default values. I'm happy to merge, and happy to leave it with you if you already have the idea in mind.

dalonsoa (approved):
To be honest, nothing to add that @alexdewar has not explained much, much better - and with examples.

PR #199, opened 2023-08-08
Load nutrients

### LINKED ISSUE #162: Populate food nutrient coefficients in France.NewEBM.json
Some food to nutrient coefficient data has been provided as an Excel file: 

https://imperiallondon-my.sharepoint.com/:x:/g/personal/jpturner_ic_ac_uk/EewKQrZoodhOkG1r7ealmmIB2wZtgGSMBXmWucJiD3yy-A?e=l9SHL5

We should populate the `Food` section of `France.NewEBM.json` with these values, which would probably involve implementing a script (Python?)  to automate this process (it's quite long). See #133

jamesturner246:
Fixes #162 

I'll get this going, as the bulk of it is there.

I have a new example directory `example_new`, containing the new configs. France.NewEBM.json has another field: `FoodsDataFile`, specifying the CSV foods nutrients file similar to existing syntax in main config file./

the data is loaded inside model_parser `load_newebm_model`, and for now that's it. I did a little tidy up in csvparser as well. The helper functions are put in an anonymous namespace in `csvparser.cpp`, and declarations are removed from the header. This should make the interfaces a little less cluttered with helper fns.

I'll finalise and merge this after #200 merges, since it uses a bit of common code, which is now changing.


src/HealthGPS.Console/csvparser.cpp
```diff
@@ -95,85 +97,86 @@ create_fields_index_mapping(const std::vector<std::string> &column_names,
...
     // Validate columns and create file columns map
-    auto mismatch = 0;
     auto headers = doc.GetColumnNames();
-    std::map<std::string, std::string, hc::case_insensitive::comparator> csv_cols;
-    for (auto &pair : columns) {
-        auto is_match = [&pair](const auto &str) {
-            return hc::case_insensitive::equals(pair.first, str);
+    std::map<std::string, std::string, hc::case_insensitive::comparator> csv_column_map;
+    for (const auto &[col_name, col_type] : columns) {
+        auto is_match = [&col_name](const auto &csv_col_name) {
```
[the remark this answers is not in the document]
jamesturner246: Eh?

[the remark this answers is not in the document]
jamesturner246: This seems to be a bug in clang, where it can't capture structured bindings in lambdas, even though C++20 allows it. https://stackoverflow.com/questions/46114214/lambda-implicit-capture-fails-with-variable-declared-from-structured-binding

Work to fix it here: https://reviews.llvm.org/D122768

For now I guess we don't use structured bindings here, unless we can work around that lambda.

src/HealthGPS.Console/csvparser.cpp
```diff
@@ -95,85 +97,86 @@ create_fields_index_mapping(const std::vector<std::string> &column_names,
...
+            return hc::case_insensitive::equals(col_name, csv_col_name);
```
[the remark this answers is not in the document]
jamesturner246: As above.

example_new/MAFFCODE with NUTRIENTS and NPM scores.xlsx
dalonsoa, line 1: Why an Excel file? Is this really necessary and cannot be done with CSV files? Excel files are really inconvenient for version control.

src/HealthGPS.Console/configuration.cpp
```diff
@@ -154,25 +156,26 @@ Configuration load_configuration(CommandOptions &options) {
...
-        if (full_path.is_relative()) {
-            full_path = options.config_file.parent_path() / config.file.name;
-            if (fs::exists(full_path)) {
-                config.file.name = full_path.string();
-                fmt::print("Input dataset file..: {}\n", config.file.name);
-            }
+        file_path = config.file.name;
+        if (file_path.is_relative()) {
+            file_path = config.root_path / file_path;
+            config.file.name = file_path.string();
```
dalonsoa, line 172: This same structure to make the path absolute is used in several places. I would suggest to write a small function that does the job and that you can reuse.

src/HealthGPS.Console/configuration.cpp
```diff
@@ -181,32 +184,33 @@ Configuration load_configuration(CommandOptions &options) {
...
-            if (full_path.is_relative()) {
-                full_path = options.config_file.parent_path() / model.second;
-                if (fs::exists(full_path)) {
-                    model.second = full_path.string();
-                    fmt::print("Model: {:<7}, file: {}\n", model.first, model.second);
-                }
+            file_path = model.second;
+            if (file_path.is_relative()) {
+                file_path = config.root_path / file_path;
+                model.second = file_path.string();
```
dalonsoa, line 191: As above

src/HealthGPS.Console/configuration.cpp
```diff
@@ -181,32 +184,33 @@ Configuration load_configuration(CommandOptions &options) {
...
-                item.second = full_path.string();
-                fmt::print("{:<14}, file: {}\n", item.first, full_path.string());
-            } else {
-                fmt::print(fg(fmt::color::red), "Adjustment type: {}, file: {} not found.\n",
-                           item.first, full_path.string());
+            file_path = item.second;
+            if (file_path.is_relative()) {
+                file_path = config.root_path / file_path;
+                item.second = file_path.string();
+            }
```
dalonsoa, line 206: As above

dalonsoa (approved):
It looks good. I've just made a comment on reusability of code blocks and another one on using Excel files.

example_new/MAFFCODE with NUTRIENTS and NPM scores.xlsx
jamesturner246: Indeed, this is just me being lazy until full and final data comes in. ^^ I'll remove it.

example_new/MAFFCODE_NUTRIENTS_NPM.csv
alexdewar, line 1: I'm not entirely clear on what the difference between the current `example` and `data` folders is, but might this belong in a data folder instead? The other one has things like country codes in there.

src/HealthGPS.Console/csvparser.cpp
```diff
@@ -79,101 +81,91 @@ hc::DoubleDataTableColumnBuilder parse_double_column(std::string name,
...
+            success = false;
         }
     }
 
-    return result;
+    if (!success) {
+        throw std::runtime_error("Error parsing dataset.");
+    }
+
+    return success;
```
jamesturner246, line 150: I should really just return `out_table` instead of the `success` bool, now that I have exceptions. Doing so is not entirely trivial, though, as the default move constructor for the table cannot be constructed, possibly due to the mutex. It's possible we can just manually define the move constructor, which will `std::move(mutex)`. Not sure. Needs trying.

alexdewar (changes requested):
The code changes look sensible to me. I've made some minor suggestions.

I'm not entirely clear what the contents of the `example_new` folder are though, other than that they're intended to be ingested by the new model. How do the files differ from those in `example/` if at all? Are they just going to be example data? Or were you intending to replace them with real-world data at some point? Real-world data should probably live somewhere outside this repo; example data's fine, but could we possibly use a smaller dataset? I know the existing example dataset is huge, but I don't know that we want to commit 100,000s of lines of data to the repo every time we add a new model. Maybe `example_new` needs a readme to explain things a bit?

I think we should probably merge this before #200, because a) it'll be ready sooner and b) the refactoring on #200 is more extensive, so I think it's easier to do it this way round. In future we probs need to coordinate a bit better because there are going to be a whole lot of merge conflicts here (I'm not saying this is your fault!).

example_new/MAFFCODE with NUTRIENTS and NPM scores.xlsx
jamesturner246: Yes, for now I think. Once they start analysing it and deciding which bits they do and don't want, then I'll remove delete as required.

example_new/France.HLM.json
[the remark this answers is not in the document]
jamesturner246: Yes. Temporary until the new initialisation is worked out. The dynamic model name is different: `NewEBM`, but that's it.

example_new/MAFFCODE_NUTRIENTS_NPM.csv
jamesturner246: It's all quite temporary for the moment, until data and the rest of the model comes in, and we can organise it properly.

jamesturner246:
The `examples_new` is quite temporary. It's using parts of the older model, until the new model and data come in. I made a new dir temporarily so I can change things more freely without breaking the old one. Given the state of flux, I don't think we should prematurely optimise these until we actually know what's going in them.

Agree on the coordination point .Understand it's a bit annoying. it's quite hard to change things at the moment, given how tightly coupled the various `config`, `settings`, `options`, `poco` things are.

example_new/MAFFCODE_NUTRIENTS_NPM.csv
jamesturner246: A better explanation: I wasn't really sure about this myself at first.

The `data` dir contains more permanent and widely applicable data, such as country codes (don't change often), diseases (also don't), etcetera.

The `example` dir contains input data for the current experiment, such as population initialisation and extrapolation data for France, at year 20xx. 

There's probably some improvements to be made here, but, for now at least, there's separation of generally common and static data, and the data for the current experiment (year, country, etcetera). Hope that makes more sense.

src/HealthGPS.Console/csvparser.cpp
jamesturner246: Are there any consequences of wrapping mutexes in unique pointers? Besides still not having copy ctors?

example_new/MAFFCODE_NUTRIENTS_NPM.csv
```diff
@@ -0,0 +1,513 @@
...
+300101,Fully-coated chocolate biscuits or wafers,26,16.18,7.58,0.69,63,500,0.12,50.1,42.5,0,0,12,7.5,24
+300102,Sweet biscuits including half-coated chocolate biscuits,23.4,7.46,9.99,1.98,62.2,469,0.36,24.1,23.8,0,1.5,22,6.2,12
+300103,Cereal bars and cereal based cakes,22.2,4.48,11.37,5.35,60.5,468,0.074,32.3,28,0,4.6,0,10.4,12
+300104,Savoury biscuits,16.3,5.77,4.38,5.32,68.3,440,0.61,0,0,0,6.1,0,9.5,6
+310101,"Nuts, nut products and seeds",53,9.5,24.2,16.5,7.1,602,0.4,3.8,0,0,6.9,0,24.5,14
+310102,"Potato crisps or snacks including unspecified 'crisps', prawn crackers",34.2,14.03,13.68,4.97,53.3,530,0.84,0.7,0,0,11.6,0,5.7,18
+310103,"Cornsnacks, based on maize",31.9,11.77,12.91,5.78,59.3,537,1.13,4.6,0,0,2.1,0,7,21
+310104,Wheat based savoury snack,28,17.24,7.71,1.76,51.8,488,0.935,2.2,0,0,5.7,46,10.4,5
+310201,Popcorn,20,1.72,9.48,7.01,77.6,480,0.056,62.1,61.8,0,1.8,18,2.1,15
+310301,Other savoury snacks (including hors d'oeuvres),16.246,5.924,7.152,1.933,13.365,244.639,0.402103,0.287,0.012,0,0.53,57.59,12.12,0
```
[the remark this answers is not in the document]
jamesturner246: I hope they put a lot of thought into it!  xD

src/HealthGPS.Console/csvparser.cpp
alexdewar: Not really. You get one extra heap allocation of a few bytes when you create the object, but I don't think we have to worry about that too much :stuck_out_tongue:

src/HealthGPS.Console/csvparser.cpp
```diff
@@ -116,6 +116,7 @@ bool load_datatable_from_csv(hgps::core::DataTable &out_table, const poco::FileI
         throw std::runtime_error("Required columns not found in dataset.");
     }
 
+    hgps::core::DataTable out_table{};
```
alexdewar, line 119: FYI `DataTable` should be default initialised here anyway. You don't need the extra braces.

alexdewar (approved):
LGTM!

PR #213, opened 2023-09-20
More pre-commit hooks for formatting

alexdewar:
I've added a pre-commit hook for formatting CMake files, as some of them were a bit messy (tabs, eww). I also added a couple of hooks for removing trailing whitespace and ensuring files end with a `\n`.

I also deleted a bunch of unneeded VS files.


src/HealthGPS.Console/versioninfo.rc
jamesturner246, line 1: What even is this?

jamesturner246 (approved):
LGTM. One less OCD trigger.

src/HealthGPS.Console/versioninfo.rc
alexdewar: It's needed by Visual Studio, but I'm not sure what for!

PR #214, opened 2023-09-22
Food initialisation

### LINKED ISSUE #207: Load foods correlation matrix and decompose it

### LINKED ISSUE #208: Getting the residuals

jamesturner246:
This PR adds a new `StaticLinearModel` class, with correlated residual sampling, to be used in conjunction with the new Kevin Hall energy balance dynamic model. The purpose of the model is to initialise risk factor (food) values using a linear model, and introduce stochasticity with random (correlated) residuals. Its advantage over the existing hierarchical linear model, besides its simplicity, is that it samples residuals by sampling correlated Gaussian models, rather than saving hundreds of real residuals in memory.

A simple linear model is used to approximate risk factor values by age and sex, whose values are now drawn from the new static model config file: `./example_new/StaticLinear.json`.

For stochasticity, a Cholesky decomposition matrix is saved inside this model, and random zero-mean unit-variance numbers are sampled with the existing HealthGPS RNG, which are multiplied with the Cholesky decomposition using the Eigen library.

The main parts of the change reside in `static_linear_model.cpp`, and associated `model_parser.cpp` code. The new config directory is also cleaned up a bunch.

The existing 'baseline adjustment' mechanism is used to bring the distribution of simulated risk factors in line with real mean by age and sex factor data, now present in `./example_new/FactorsMean.*.csv`.

NB: this class is incomplete (waiting for info). The risk factors do not have their residuals added yet.

Fixes #207 

Fixes #208


src/HealthGPS/static_linear_model.cpp
```diff
@@ -0,0 +1,118 @@
...
+        auto samples = correlated_samples(context);
+    }
+}
+
+Eigen::VectorXd StaticLinearModel::correlated_samples(RuntimeContext &context) {
+
+    // Correlated samples using Cholesky decomposition.
+    Eigen::VectorXd samples{risk_factor_names_.size()};
+    for (size_t i = 0; i < risk_factor_names_.size(); i++) {
+        samples[i] = context.random().next_normal(0.0, 1.0);
```
dalonsoa, line 66: Just to confirm I get this right, this generates random data following a random distribution with mean 0 and standard deviation 1, right? It does not generate random data between 0 and 1, correct?

src/HealthGPS/static_linear_model.cpp
```diff
@@ -0,0 +1,118 @@
...
+    }
+    samples = risk_factor_cholesky_ * samples;
+
+    // Bound and scale the samples.
+    double sigma_cap = 3.0; // for removing outliers
+    for (size_t i = 0; i < risk_factor_names_.size(); i++) {
+        samples[i] = std::min(std::max(samples[i], -sigma_cap), sigma_cap);
+        samples[i] = samples[i] / sigma_cap;
```
dalonsoa, line 74: Why do you need to divide by `sigma_cap`?

src/HealthGPS/static_linear_model.cpp
```diff
@@ -0,0 +1,118 @@
...
+    if (risk_factor_names_.empty()) {
+        throw core::HgpsException("Risk factor names list is empty");
+    }
+    if (risk_factor_models_.intercepts.empty() || risk_factor_models_.coefficients.empty()) {
+        throw core::HgpsException("Risk factor models mapping is incomplete");
+    }
+    if (!risk_factor_cholesky_.allFinite()) {
+        throw core::HgpsException("Risk factor Cholesky matrix contains non-finite values");
+    }
+}
```
dalonsoa, line 111: What's the difference between this and the `StaticLinearModel` above?

src/HealthGPS/static_linear_model.h
```diff
@@ -0,0 +1,68 @@
...
+/// @details The static model is used to initialise the virtual population.
+class StaticLinearModel final : public HierarchicalLinearModel {
+  public:
+    /// @brief Initialises a new instance of the StaticLinearModel class
+    /// @param risk_factor_names An ordered list of risk factor names
+    /// @param risk_factor_models The linear models used to initialise a person's risk factor values
+    /// @param risk_factor_cholesky The Cholesky decomposition of the risk factor correlation matrix
+    StaticLinearModel(std::vector<core::Identifier> risk_factor_names,
+                      LinearModelParams risk_factor_models, Eigen::MatrixXd risk_factor_cholesky);
+
```
dalonsoa, line 27: This also throws and exception if the arguments are invalid, right?

dalonsoa (approved):
Looks good to me!

Given that we are skipping adding test, for now, given time constrains, it might be worth opening an issue to remind ourselves about adding tests to all of these functions, one day.

src/HealthGPS/static_linear_model.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+Eigen::VectorXd StaticLinearModel::correlated_samples(RuntimeContext &context) {
+
+    // Correlated samples using Cholesky decomposition.
+    Eigen::VectorXd samples{risk_factor_names_.size()};
+    for (size_t i = 0; i < risk_factor_names_.size(); i++) {
+        samples[i] = context.random().next_normal(0.0, 1.0);
+    }
```
alexdewar, line 67: If you `#include <ranges>` this can be written as:

```suggestion
    std::ranges::generate(samples, [&context] { return context.random().next_normal(0.0, 1.0); });
```

src/HealthGPS/static_linear_model.cpp
```diff
@@ -0,0 +1,118 @@
...
+    samples = risk_factor_cholesky_ * samples;
+
+    // Bound and scale the samples.
+    double sigma_cap = 3.0; // for removing outliers
+    for (size_t i = 0; i < risk_factor_names_.size(); i++) {
+        samples[i] = std::min(std::max(samples[i], -sigma_cap), sigma_cap);
+        samples[i] = samples[i] / sigma_cap;
+    }
```
alexdewar, line 75: Similarly, this can be:

```suggestion
    std::ranges::transform(samples, samples.begin(), [](double sample) {
        constexpr double sigma_cap = 3.0; // for removing outliers
        return std::clamp(sample, -sigma_cap, sigma_cap) / sigma_cap;
    });
```

vcpkg.json
```diff
@@ -5,6 +5,7 @@
     "dependencies": [
         "fmt",
         "cxxopts",
+        "eigen3",
```
alexdewar, line 8: You should also add this to the list of dependencies in the README.

alexdewar (changes requested):
I've made minor suggestions, but otherwise all good!

src/HealthGPS/static_linear_model.cpp
jamesturner246: Correct. Args are mean and std dev.

src/HealthGPS/static_linear_model.cpp
jamesturner246: The reason is still not entirely clear to me, but what this is doing is filtering +/- 3 stddev outliers, and then scaling the distribution to +/- 1, hence the division by the cap. This will later be multiplied by some fraction of the (baseline) mean, producing the residual.

src/HealthGPS/static_linear_model.cpp
jamesturner246: This is the 'definition' class which generates an actual model instance with its `create_model` method, due to the factory design pattern and code structure that currently exists. The idea is we want to be able to initialise a fresh new model instance on-demand, in case some GUI requests a simulation reset.

It's not ideal, and the whole thing should and hopefully will be refactored given chance in future.

src/HealthGPS/static_linear_model.h
jamesturner246: Correct. I'll add the throws docstring.

src/HealthGPS/static_linear_model.cpp
```diff
@@ -0,0 +1,118 @@
...
+        // Correlated residual sampling.
+        auto samples = correlated_samples(context);
+    }
+}
+
+void StaticLinearModel::update_risk_factors(RuntimeContext &context) {
+
+    for (auto &person : context.population()) {
+
+        // Only newborns should be initialised.
```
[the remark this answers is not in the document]
jamesturner246: Nope. The code should initialise newborns (age == 0) and skip initialisation (continue) if not newborn (age > 0). Perhaps confusingly, what follows this check is initialisation code (even though inside an update function).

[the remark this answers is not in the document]
jamesturner246: Maybe I'll change the comment to "Skip initialisation if not newborn".

src/HealthGPS/static_linear_model.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+        // Approximate risk factor values with linear models.
+        linear_approximation(person);
+
+        // Correlated residual sampling.
+        auto samples = correlated_samples(context);
+    }
+}
+
+void StaticLinearModel::update_risk_factors(RuntimeContext &context) {
```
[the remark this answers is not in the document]
jamesturner246: Not in this case, as `update_risk_factors` is an override, where other models, including this one eventually, will need write access to the context.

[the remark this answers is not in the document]
jamesturner246: ... actually, that might not be true. Right now they just need write access to the `Population` contained in context. I'll open an issue for later, since it requires changing the RF model interface.

alexdewar (approved):
LGTM!

PR #219, opened 2023-09-27
Add code for loading nutrient table

### LINKED ISSUE #217: Fix up nutrient loading code

alexdewar:
Add code for loading nutrient table from a CSV file in the form of a `load_nutrient_table()` function. This is currently not integrated with the rest of the code, though I have added tests to verify that it works. I figured that the integration might be dependent on what @jamesturner246 is working on, so it was better to merge this first, then worry about the integration later.

I also added a user-defined literal for `Identifier`s so you can now write `"Disease"_id` etc. (instead of `Identifier{"Disease"}`).

Fixes #217.

### DIFF
--- a/src/HealthGPS.Tests/LoadNutrientTable.cpp
+++ b/src/HealthGPS.Tests/LoadNutrientTable.cpp
@@ -10,17 +10,10 @@
 using namespace hgps;
 
 namespace {
-constexpr auto *csv_data = R"(food,d,e,a,c,b
-chicken,1,2,3,4,5
-crisps,6,7,8,9,10
-tofu,11,12,13,14,15
-)";
-
 const std::vector<core::Identifier> nutrients = {"A"_id, "B"_id, "C"_id};
 } // anonymous namespace
 
 TEST(LoadNutrientTable, GetNutrientIndexes) {
-
     {
         // Note that the case doesn't match; should be case insensitive
         const std::vector<std::string> column_names = {"d", "e", "a", "c", "b"};
@@ -35,16 +28,6 @@ TEST(LoadNutrientTable, GetNutrientIndexes) {
         const std::vector<std::string> column_names = {"D", "E", "A", "C"};
         EXPECT_THROW(detail::get_nutrient_indexes(column_names, nutrients), core::HgpsException);
     }
-}
-
-TEST(LoadNutrientTable, GetNutrientTable) {
-    std::istringstream is{csv_data};
-    rapidcsv::Document doc{is, rapidcsv::LabelParams{0, 0}};
-    const auto table = detail::get_nutrient_table(doc, nutrients, {2, 4, 3});
-
-    const decltype(table) expected = {
-        {"chicken", {3, 5, 4}}, {"crisps", {8, 10, 9}}, {"tofu", {13, 15, 14}}};
-    EXPECT_EQ(table, expected);
 }
 
 TEST(LoadNutrientTable, LoadNutrientTable) {
--- a/src/HealthGPS.Tests/LoadNutrientTable.cpp
+++ b/src/HealthGPS.Tests/LoadNutrientTable.cpp
@@ -1,0 +1,57 @@
+#include "pch.h"
+#include "program_path.h"
+
+#include "HealthGPS.Core/exception.h"
+#include "HealthGPS/nutrients.h"
+
+#include <filesystem>
+#include <sstream>
+
+using namespace hgps;
+
+namespace {
+constexpr auto *csv_data = R"(food,d,e,a,c,b
+chicken,1,2,3,4,5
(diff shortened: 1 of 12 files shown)


src/HealthGPS.Tests/LoadNutrientTable.cpp
```diff
@@ -0,0 +1,57 @@
...
+#include <sstream>
+
+using namespace hgps;
+
+namespace {
+constexpr auto *csv_data = R"(food,d,e,a,c,b
+chicken,1,2,3,4,5
+crisps,6,7,8,9,10
+tofu,11,12,13,14,15
+)";
```
dalonsoa, line 17: Why do you need this if you already have the csv data file below for tests?

src/HealthGPS.Tests/program_path.cpp
dalonsoa, line 1: Why are these checks about the platform necessary now and not before? Why your new functionality requires them?

src/HealthGPS/CMakeLists.txt
dalonsoa, line 1: I'm curious about this. Is there no "autodiscovery" feature in CMake that facilitate putting together this list of target sources? Having to manually add every file seems so brittle and a potential source of errors if you forget to include one or change its name or something like that...

dalonsoa (approved):
Looks good! I just had a few questions.

src/HealthGPS.Tests/LoadNutrientTable.cpp
alexdewar: Hmm... Good point. I wrote this test first, because reading from a file in your unit tests involves faffing around with CMake, but then I ended up doing that anyway. So we could just drop this.

src/HealthGPS/CMakeLists.txt
alexdewar: There's not a *good* way to do it, though you could just get CMake to find every `*.cpp` file in the current folder and build them.

The reason you don't want to do this is because you need to reconfigure the project with CMake when files are added/removed and the only way the generated `Makefile` (for example) knows to do this is by checking if `CMakeLists.txt` has changed (e.g. because you've done a `git pull`). So it's basically a design limitation of CMake -- I think there are other build systems that are smarter than this.

alexdewar: I just googled this because I was curious and it turns out they did add this feature to CMake at some point! My knowledge is clearly outdated :laughing:

See: https://stackoverflow.com/questions/3201154/automatically-add-all-files-in-a-folder-to-a-target-using-cmake

alexdewar: ...although seemingly the CMake developers still recommend you do things the old way: https://cmake.org/cmake/help/latest/command/file.html#filesystem

Ahh, the joys of working with C++.

src/HealthGPS/nutrients.h
```diff
@@ -0,0 +1,30 @@
...
+#include "HealthGPS.Core/identifier.h"
+
+#include "rapidcsv.h"
+
+#include <unordered_map>
+#include <vector>
+
+namespace hgps {
+namespace detail {
+/// @brief Work out which column each macronutrient isin
```
jamesturner246, line 12:
```suggestion
/// @brief Work out which column each macronutrient is in
```

src/HealthGPS/nutrients.h
```diff
@@ -0,0 +1,30 @@
...
+/// @param column_names Column names for CSV file
+/// @param nutrients Macronutrients
+/// @return Index for each macronutrient in order
+std::vector<size_t> get_nutrient_indexes(const std::vector<std::string> &column_names,
+                                         const std::vector<core::Identifier> &nutrients);
+
+std::unordered_map<core::Identifier, std::vector<double>>
+get_nutrient_table(rapidcsv::Document &doc, const std::vector<core::Identifier> &nutrients,
+                   const std::vector<size_t> &nutrient_idx);
```
jamesturner246, line 21: Are these supposed to be internal functions used by `load_nutrient_table`? If so, would it be best to leave them out of the header?

jamesturner246 (approved):
Approve in general. I liked the overload quote trick! Just a couple of things before merging.

src/HealthGPS/nutrients.h
alexdewar: Yep. The only thing is that I have a unit test for `get_nutrient_indexes()`, so I need some way of accessing it. I suppose one way to do it would be to put the function declaration in the test file itself, though that's maybe a bit unorthodox. What do you think?

jamesturner246: Ah, good point. I used to have a separate header for internal functions, but that was C. I'd say it's fine as is then.

PR #225, opened 2023-10-17
Class and file renaming

### LINKED ISSUE #223: Rename misleadingly named `HierarchicalLinearModel` interface
`HierarchicalLinearModel` is the interface to all the static/dynamic risk factor models, and not necessarily just the HLM. This has tripped me up on more than one occasion already. Its name should change to reflect this.

- [ ] renaming HierarchicalLinearModel to RiskFactorModel, or similar.
- [ ] renaming HierarchicalModelType to RiskFactorModelType, or similar.

jamesturner246:
A sorely needed renaming refactor. This may look big, but all that's really going on is renaming classes and files to better reflect what they do.

Fixes #223. 

In summary, the changes are as follows:

RENAMED CLASSES AND FILES

* rename HierarchicalLinearModel to RiskFactorModel
* rename HierarchicalModelType to RiskFactorModelType

* rename energy_balance_hierarchical_model.* to dynamic_hierarchical_linear_model.*
* rename EnergyBalanceHierarchicalModel to DynamicHierarchicalLinearModel
* rename LiteHierarchicalModelDefinition to DynamicHierarchicalLinearModelDefinition

* rename energy_balance_model.* to kevin_hall_model.*
* rename EnergyBalanceModel to KevinHallModel
* rename EnergyBalanceModelDefinition to KevinHallModelDefinition

* rename hierarchical_model_static.* to static_hierarchical_linear_model.*
* rename HierarchicalLinearModelDefinition to StaticHierarchicalLinearModelDefinition:

HEADER CLEANING

* remove hierarchical_model.h (unused header file which includes other headers)

* move repository.h risk factor model includes to model_parser.h


src/HealthGPS/dynamic_hierarchical_linear_model.h
```diff
@@ -74,21 +73,21 @@ class EnergyBalanceHierarchicalModel final : public HierarchicalLinearModel {
 };
 
 /// @brief Defines the lite hierarchical linear model data type
```
dalonsoa, line 75: Maybe this docstrings needs updating, too?

dalonsoa (approved):
Wow! I don't know if the names now are more appropriate but things now look consolidated and somewhat related with each other.

Obviously not for this PR, but is there anywhere in the documentation stating the models that are available and describing them, so the user can choose what to use? If not, I would make an issue of it to be tackled in the future. It is good to have different options and properly named, but if they are not explained anywhere, they will be hard to use, regardless!

src/HealthGPS/dynamic_hierarchical_linear_model.h
jamesturner246: Ah, good spot, thanks!

PR #230, opened 2023-10-20
Sector and income

### LINKED ISSUE #228: Initialise a Person's sector (region)
This will involve loading sector (region) prevalence for age groups and sex in via model parser into the Kevin Hall model, and updating `generate_risk_factors` to initialise a `Person`s sector.

### LINKED ISSUE #229: Initialise a Person's income
This will involve loading another simple linear model in from the Kevin Hall config JSON into the `KevinHall` class, and modifying `generate_risk_factors` to initialise income for a `Person`.

This is not the ideal place to put it, and the KH model is becoming a little cluttered. Ideally this would go inside a new socio-econiomic standing (SES) model. However, client time constraints... 

This should be revisited.

See also #228

jamesturner246:
In this PR, we add two models for initialising and updating a `Person`'s sector (region) and income level category. The changes are summarised as follows:

* Adds `generate` and `update` methods to `KevinHallModel` for sector and income. This is not the ideal place for them, as this class is becoming bloated. In My future refactor plans, these both would become their own 'modules', where the user has control of number of modules and ordering. 

* Allows both `generate` and `update` methods of risk factor models to always run -- see #231.

* Adds new fields and accessors to `Person`, and some error checking for unknown data.

* A general cleanup as best as possible for now

Fixes #228

Fixes #229


src/HealthGPS/kevin_hall_model.cpp
```diff
@@ -114,6 +139,89 @@ void KevinHallModel::update_risk_factors(RuntimeContext &context) {
...
 
+void KevinHallModel::initialise_sector(RuntimeContext &context, Person &person) const {
+
+    // Get rural prevalence for age group and sex.
+    double prevalence;
+    if (person.age < 18) {
+        prevalence = rural_prevalence_.at("Under18"_id).at(person.gender);
+    } else {
+        prevalence = rural_prevalence_.at("Over18"_id).at(person.gender);
+    }
```
alexdewar, line 150: Do you need to have the proportion of both over- and under-18s defined in the config file? If you just define one of them, couldn't you calculate the other as `1 - x`?

src/HealthGPS/kevin_hall_model.cpp
```diff
@@ -114,6 +139,89 @@ void KevinHallModel::update_risk_factors(RuntimeContext &context) {
...
+
+    // Sample the person's sector.
+    double rand = context.random().next_double();
+    auto sector = rand < prevalence ? core::Sector::rural : core::Sector::urban;
+    person.sector = sector;
```
alexdewar, line 155: This is a perfectly fine way of setting `sector` randomly, but I just want to double-check that the model is designed this way on purpose...

Another way to do it would be to say e.g. for every 100 people, we want 60 assigned as rural and 40 as urban, then shuffle the order. That way you would always have the exact proportion as defined in the config file.

Out of interest, how are other attributes (e.g. gender) assigned randomly?

src/HealthGPS/kevin_hall_model.cpp
```diff
@@ -114,6 +139,89 @@ void KevinHallModel::update_risk_factors(RuntimeContext &context) {
...
+}
+
+void KevinHallModel::update_sector(RuntimeContext &context, Person &person) const {
+
+    // Only update rural sector 18 year olds.
```
alexdewar, line 160: How come?

alexdewar: Also, why are you looking at people who are *exactly* 18, rather than over 18?

alexdewar: Ohhh I see. You're assuming that migration to the city always happens at age 18. I think I remember Ali talking about this.

src/HealthGPS/kevin_hall_model.cpp
```diff
@@ -114,6 +139,89 @@ void KevinHallModel::update_risk_factors(RuntimeContext &context) {
...
+void KevinHallModel::initialise_income(RuntimeContext &context, Person &person) const {
+
+    // Compute logits for each income category.
+    auto logits = std::vector<double>(income_models_.size());
+    for (size_t i = 0; i < income_models_.size(); i++) {
+        logits[i] = income_models_[i].intercept;
+        for (const auto &[factor_name, coefficient] : income_models_[i].coefficients) {
+            logits[i] += coefficient * person.get_risk_factor_value(factor_name);
+        }
+    }
```
alexdewar, line 186: Alternative implementation using ranges (you'll also need `#include <ranges>`):

```suggestion
    auto logits = std::vector<double>{};
    logits.reserve(income_models_.size());
    std::ranges::transform(income_models_, std::back_inserter(logits),
                           [&person](const auto &model) {
                               double logit = model.intercept;
                               for (const auto &[factor_name, coefficient] : model.coefficients) {
                                   logit += coefficient * person.get_risk_factor_value(factor_name);
                               }
                               return logit;
                           });
```

src/HealthGPS/kevin_hall_model.cpp
```diff
@@ -114,6 +139,89 @@ void KevinHallModel::update_risk_factors(RuntimeContext &context) {
...
+    for (size_t i = 0; i < income_models_.size(); i++) {
+        e_logits[i] = exp(logits[i]);
+        e_logits_sum += e_logits[i];
+    }
+
+    // Compute income category probabilities.
+    auto probabilities = std::vector<double>(income_models_.size());
+    for (size_t i = 0; i < income_models_.size(); i++) {
+        probabilities[i] = e_logits[i] / e_logits_sum;
+    }
```
alexdewar, line 200: These could also use `std::ranges::transform`, though the way you've done it is fine too!

One thing I would say is that generally it's better to append to `std::vector`s iteratively rather than creating them with a particular size upfront (it means you can skip the zeroing of the memory). You can use the `reserve()` method to hint at how many elements it will have eventually.

src/HealthGPS/kevin_hall_model.cpp
```diff
@@ -114,6 +139,89 @@ void KevinHallModel::update_risk_factors(RuntimeContext &context) {
...
+}
+
+void KevinHallModel::update_income(RuntimeContext &context, Person &person) const {
+
+    // Only update 18 year olds.
+    if (person.age != 18) {
+        return;
+    }
+
+    initialise_income(context, person);
```
alexdewar, line 222:
```suggestion
    if (person.age == 18) {
        initialise_income(context, person);
    }
```

src/HealthGPS/person.cpp
```diff
@@ -56,14 +60,29 @@ double Person::get_risk_factor_value(const core::Identifier &key) const {
...
+        throw core::HgpsException("Gender is unknown.");
+    }
     return gender == core::Gender::male ? "male" : "female";
 }
 
+float Person::sector_to_value() const {
+    if (sector == core::Sector::unknown) {
+        throw core::HgpsException("Sector is unknown.");
+    }
+    return sector == core::Sector::urban ? 0.0f : 1.0f;
```
alexdewar, line 81: Given that it's a bit ambiguous as to what 0 and 1 mean in this context, do we maybe want to rename the sector property to "is_urban" or something?

I suppose the downside of that is that it wouldn't be particularly extensible if more sectors are added in future.

alexdewar (approved):
I've got a few queries and small suggestions, but otherwise this all looks pretty sensible!

example_new/KevinHall.json
```diff
@@ -67,6 +67,43 @@
...
+            "Name": "Under18",
+            "Female": 0.65,
+            "Male": 0.65
+        },
+        {
+            "Name": "Over18",
+            "Female": 0.6,
+            "Male": 0.6
+        }
+    ],
```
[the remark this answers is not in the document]
jamesturner246: I understand the confusion. What it's saying is that, for example, in the group of women aged over 18, 60% of them are from the rural sector. But yeah, it's a bit misleading.

src/HealthGPS/dynamic_hierarchical_linear_model.cpp
```diff
@@ -26,10 +26,7 @@ RiskFactorModelType DynamicHierarchicalLinearModel::type() const noexcept {
 std::string DynamicHierarchicalLinearModel::name() const noexcept { return "Dynamic"; }
 
 void DynamicHierarchicalLinearModel::generate_risk_factors(
-    [[maybe_unused]] RuntimeContext &context) {
-    throw core::HgpsException(
-        "DynamicHierarchicalLinearModel::generate_risk_factors not yet implemented.");
-}
+    [[maybe_unused]] RuntimeContext &context) {}
```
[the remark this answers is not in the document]
jamesturner246: Doesn't generate risk factors. That's done in e.g. `StaticHierarchicalLinearModel`. Alas, the confusion of using the generate (static) model and update (dynamic) model terminology when both models have generate and update methods. That's why I will do #231.

src/HealthGPS.Console/model_parser.cpp
```diff
@@ -272,6 +287,9 @@ load_kevinhall_risk_model_definition(const poco::json &opt, const host::Configur
     }
 
     // Food groups.
+    std::unordered_map<hgps::core::Identifier, std::map<hgps::core::Identifier, double>>
+        nutrient_equations;
+    std::unordered_map<hgps::core::Identifier, std::optional<double>> food_prices;
```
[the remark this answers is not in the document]
jamesturner246: True, I didn't do that as I didn't think much arithmetic would happen with them. They aren't used yet, and will presumably be just logged for financial impact of interventions. They can be changed to unsigned integral types for pence/cents/... I'll leave that for another time, if and as needed.

src/HealthGPS.Console/model_parser.cpp
```diff
@@ -253,14 +272,10 @@ load_ebhlm_risk_model_definition(const poco::json &opt) {
...
-    std::unordered_map<hgps::core::Identifier, double> energy_equation;
-    std::unordered_map<hgps::core::Identifier, std::pair<double, double>> nutrient_ranges;
-    std::unordered_map<hgps::core::Identifier, std::map<hgps::core::Identifier, double>>
-        nutrient_equations;
-    std::unordered_map<hgps::core::Identifier, std::optional<double>> food_prices;
-    std::unordered_map<hgps::core::Gender, std::vector<double>> age_mean_height;
 
     // Nutrient groups.
+    std::unordered_map<hgps::core::Identifier, double> energy_equation;
+    std::unordered_map<hgps::core::Identifier, std::pair<double, double>> nutrient_ranges;
```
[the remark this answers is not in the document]
jamesturner246: Good spot. It would indeed.

src/HealthGPS/person.cpp
jamesturner246: Exactly that. Also the choice of rural = 1 is completely arbitrary, and based on how the coefficient data inputs were computed. Worth revisiting, but not now.

example_new/KevinHall.json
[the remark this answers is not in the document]
alexdewar: Ohhh I see. My bad for not reading it properly.

src/HealthGPS/kevin_hall_model.cpp
jamesturner246: I think the stochasticity is important to them, even though I agree that it's uncertain what purpose it is serving in this case. What you suggest is pretty much how they do it for age and sex in `PopulationModule::initialise_population`. But they don't shuffle, so I think you get bit blocks of heterogeneous people. Therefore I think the stochasticity is warranted here, so you don't get all older males living in the city and younger women living in farms. Also, since new people are periodically initialised, you would presumably need to shuffle every iteration, which sounds expensive.

alexdewar: Ah I see. If you're doing it every iteration, then it makes more sense. I think I was assuming this was done during initialisation.

src/HealthGPS/kevin_hall_model.cpp
jamesturner246: Fair enough with reserve, but I think I prefer the readability of for loops for such simple ops.

src/HealthGPS.Console/model_parser.cpp
```diff
@@ -129,31 +129,50 @@ std::unique_ptr<hgps::StaticLinearModelDefinition>
...
         host::get_file_info(opt["RiskFactorCorrelationFile"], config.root_path);
     const auto correlations_table = load_datatable_from_csv(correlations_file_info);
     Eigen::MatrixXd correlations{correlations_table.num_rows(), correlations_table.num_columns()};
-    for (size_t col = 0; col < correlations_table.num_columns(); col++) {
-        names.emplace_back(correlations_table.column(col).name());
-        for (size_t row = 0; row < correlations_table.num_rows(); row++) {
-            correlations(row, col) =
-                std::any_cast<double>(correlations_table.column(col).value(row));
+
+    for (size_t i = 0; i < opt["RiskFactorModels"].size(); i++) {
```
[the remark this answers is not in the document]
jamesturner246: Seems to be a C++23 feature :(

src/HealthGPS.Core/interval.h
```diff
@@ -33,13 +41,7 @@ template <Numerical TYPE> class Interval {
...
     /// @param value The value to check
     /// @return true if the value is in the interval; otherwise, false.
-    bool contains(TYPE value) const noexcept {
-        if (lower_ < upper_) {
-            return lower_ <= value && value <= upper_;
-        }
-
-        return lower_ >= value && value >= upper_;
-    }
+    bool contains(TYPE value) const noexcept { return lower_ <= value && value <= upper_; }
```
alexdewar, line 44: Do we want to change this behaviour? I agree that your way is saner, but I'm a little concerned that something might rely on this.

src/HealthGPS.Tests/AgeGenderTable.Test.cpp
```diff
@@ -159,7 +159,7 @@ TEST(TestHealthGPS_AgeGenderTable, CreateWithWrongRangerThrows) {
     using namespace hgps;
 
     auto negative_range = core::IntegerInterval(-1, 10);
-    auto inverted_range = core::IntegerInterval(10, 1);
+    auto inverted_range = core::IntegerInterval(1, 1);
```
alexdewar, line 162: As you've dropped support for inverted ranges, this bit of the test can be dropped.

alexdewar (approved):
Aside from one query re changing the behaviour of `Interval`, this LGTM.

src/HealthGPS.Tests/AgeGenderTable.Test.cpp
jamesturner246: That particular bit is testing for failure with zero-width intervals. The age_range is [lo, hi) -- i.e. low-inclusive, high-exclusive.

src/HealthGPS.Core/interval.h
jamesturner246: I did test the existing model, and it didn't trip the exception. I can't see anywhere that relies on non-sane intervals.

jamesturner246:
And a clang failure, presumably clang-14 from the platform. This version of clang seems to be particularly troublesome..

jamesturner246:
It's a known problem, at least.

https://github.com/actions/runner-images/issues/8659

src/HealthGPS.Tests/AgeGenderTable.Test.cpp
alexdewar: Ahh I seeeeee. That makes sense

alexdewar:
In the repo settings you can remove the clang runner from the list of merge requirements. Then you could maybe open an issue to fix it later.

alexdewar:
Another option vis-a-vis clang would be to just use the latest version, which might be a more useful test anyway.

PR #238, opened 2023-10-27
Small bugfix where immigrating people have no sector or income.

jamesturner246:
A tiny bug is fixed here, where an immigrating `Person`'s sector and income was not initialised.


dalonsoa (approved):
LGTM!

Issue #231, opened 2023-10-20
Generalise form 'static' and 'dynamic' risk factor models


jamesturner246:
For context, in existing HealthGPS speak, static means initialise and dynamic means update.

The concept of the 'static' and the 'dynamic' risk factor models is already a little cloudy.

Some Dynamic models also do initialisation, and vice versa. Both model types have `generate_` and `update_risk_factors` methods. So why not relax the restriction of having static and dynamic models, and just use the `generate` and `update` methods of both where appropriate?

This would also allow us to create as many risk factor models as we like, and stop polluting our models with hacky 'extras' where needed.

dalonsoa:
Is this related to #194?

jamesturner246:
Related, but opposite ideas. @israel-vieira was rightfully against this change because modularity of model components would decrease with such a change -- i.e. models would become more bloated. This change seeks to reverse that by having separate modules for each unit computation -- i.e. the opposite of what #194 proposes. I will close #194, to avoid confusion.

PR #241, opened 2023-10-31
Baseline adjustment and messaging controlled from within RF model

### LINKED ISSUE #239: Fine-grained control of baseline adjustment mesasaging
As it stands, the below behaviour may or may not be intended. This needs confirming with the modelling team.

Right now, baseline adjustment (and broadcasting adjustments to the intervention scenario) happens strictly after the static and dynamic risk factor models have completely finished, and is performed on all risk factors listed in `Person.risk_factors`. This may or may not be a problem, depending on whether dependent variables in the risk models depend on *unadjusted* independent  variables (like it is now) or *adjusted* variables within a single time step.

We may want to be able to control both when, and on what, these adjustments are applied by, for example, being able to call a baseline adjustment computation and broadcast procedure on upstream variables *within* a risk factor model, and choose the subset of variables on which adjustment is applied.

### LINKED ISSUE #240: Nutrient model and Physical Activity model
Implement the Box-Cox nutrient model and PA model.

Need to hack the baseline adjustment calculating and messaging mechanism to let the user control when it's called, rather than once at end of model.

jamesturner246:
It works! Huzza! :D And HealthGPS is now a lot more flexible. Apologies again for a rather large PR. I'll of course meet to go through it. Ignore the branch name -- this is to do with the baseline adjustment application and messaging to intervention.

In summary, the change moves the baseline adjustment system from the mail HGPS class (this turned out to be rather messy, hence all the scattered changes) into the risk factor classes where they're used. It allows the model classes to control both when it's used, and on which risk factors, and can be run an arbitrary number of times on arbitrary factor sets in a single timestep,

The core part of the PR is the new `RiskFactorAdjustableModel` abstract class, inheriting from the base `RiskFactorModel` class, which implements the core simulated mean code, compares against its own stored copy of risk factor expected values, applies the differences to the factors, and ships the adjustments off to the intervention.

If a risk factor model implementer requires baseline adjustment for some/all factors in their model, they would inherit from `RiskFactorAdjustableModel` instead of the base class, and call `adjust_risk_factors(context, vector_of_factor_keys);` in their `generate_risk_factors` or `update_risk_factors` functions where/when needed.

I did a bit of a tidy up whilst I was in there as well.

Happy to answer questions -- again, sorry for the big one..

Fixes #239


src/HealthGPS.Console/model_parser.cpp
```diff
@@ -161,6 +161,43 @@
...
+                fmt::format("'{}' not defined in male factor means.", model.name.to_string())};
+        }
+        if (!data.at(hgps::core::Gender::female).contains(model.name)) {
+            throw hgps::core::HgpsException{
+                fmt::format("'{}' not defined in female factor means.", model.name.to_string())};
+        }
+    }
+
+    // Write data structures.
+    auto risk_factor_means_table = hgps::FactorAdjustmentTable{std::move(data)};
```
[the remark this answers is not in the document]
jamesturner246: Eh?

src/HealthGPS/risk_factor_adjustable_model.h
```diff
@@ -0,0 +1,60 @@
...
+#include "map2d.h"
+#include "risk_factor_model.h"
+#include "runtime_context.h"
+
+#include <vector>
+
+namespace hgps {
+
+/// @brief Defines a table type for double values by sex and age
+using RiskFactorSexAgeTable = UnorderedMap2d<core::Gender, core::Identifier, std::vector<double>>;
```
dalonsoa, line 15: I see `sex` is coming here via `core::Gender`, but I don't see `age`. Is it coming via `core::Identifier`?

src/HealthGPS/risk_factor_adjustable_model.cpp
```diff
@@ -0,0 +1,170 @@
...
+    // Intervention scenario: recieve adjustments from baseline scenario.
+    else {
+        auto message = context.scenario().channel().try_receive(context.sync_timeout_millis());
+        if (!message.has_value()) {
+            throw core::HgpsException(
+                "Simulation out of sync, receive baseline adjustments message has timed out");
+        }
+
+        auto &basePtr = message.value();
+        auto *messagePrt = dynamic_cast<RiskFactorAdjustmentMessage *>(basePtr.get());
```
dalonsoa, line 62: I'm quite lost with this line. Could you clarify what's going on here?

src/HealthGPS/risk_factor_adjustable_model.cpp
```diff
@@ -0,0 +1,170 @@
...
+    auto &pop = context.population();
+    std::for_each(core::execution_policy, pop.begin(), pop.end(), [&](auto &person) {
+        if (!person.is_active()) {
+            return;
+        }
+
+        for (const auto &factor : keys) {
+            double delta = adjustments.at(person.gender, factor).at(person.age);
+            person.risk_factors.at(factor) += delta;
+        }
```
dalonsoa, line 81: So, just to confirm, the adjustments are received in a message for a certain number of factors and here we are pulling those appropriate for this person gender and age and adding them to that person risk factors, right?

src/HealthGPS/risk_factor_adjustable_model.cpp
```diff
@@ -0,0 +1,170 @@
...
+    });
+
+    // Baseline scenario: send adjustments to intervention scenario.
+    if (context.scenario().type() == ScenarioType::baseline) {
+        context.scenario().channel().send(std::make_unique<RiskFactorAdjustmentMessage>(
+            context.current_run(), context.time_now(), std::move(adjustments)));
+    }
```
dalonsoa, line 88: If this is sending the adjustments to the intervention scenario, is this not some sort of recurrent function that runs once the baseline and then, immediately afterwards, the intervention cases, but doing different things in both?

Maybe I'm missing something but this feels as it should be 3 functions:

1. Apply adjustments, whatever their origin
2. Apply adjustments from intervention scenario (calling 1 inside)
3. Apply adjustments from baseline scenario (calling `calculate_adjustments` and 2 inside).

As I said, I might be missing the actual workflow of the application. I'm not totally sure when messages are executed. If sending a message using `pypubsub` in python, things are purely sequential and syncronous, but I don't know if this works the same way.

src/HealthGPS/risk_factor_adjustable_model.cpp
```diff
@@ -0,0 +1,170 @@
...
+                }
+            }
+        }
+    }
+
+    return adjustments;
+}
+
+RiskFactorSexAgeTable
+RiskFactorAdjustableModel::calculate_simulated_mean(RuntimeContext &context,
```
dalonsoa, line 120: Out of `context`, this function only needs the age range and the population. Then, why passing the whole of the `context` object around? This probably also applies to other functions. 

Passing `context` hides what this function really depends on, which is not the whole context, but only those two specific values. It is way more explicit to pass that rather than a hold-it-all object whose internal structure might be unknown.

src/HealthGPS/dynamic_hierarchical_linear_model.cpp
```diff
@@ -25,8 +29,17 @@ RiskFactorModelType DynamicHierarchicalLinearModel::type() const noexcept {
...
+
+    // Adjust risk factors such that mean sim value matches expected value.
+    std::vector<core::Identifier> factor_keys;
+    factor_keys.reserve(variables_.size());
+    for (const auto &factor : variables_) {
+        // factor.second contains the factor name.
+        factor_keys.emplace_back(factor.second);
+    }
+    adjust_risk_factors(context, factor_keys);
+}
```
dalonsoa, line 42: Does this mean that any `DynamicHierarchicalLinearModel` will want to adjust the risk factors, always? For some reason, I was under the impression that some models will want to adjust the risk factors while others will not.

dalonsoa (changes requested):
I've reviewed only a few files that seemed the most relevant for the purpose of this PR, not every single file. I've indicated `Request changes` but it is more me asking some clarifications. 

In addition to those below, I have the following two:
- Are static models expected to have to adjust the risk factors? If no, why do they inherit from `RiskFactorAdjustableModel`? If yes, where is that adjustment called?
-  I see that dynamic models do inherit from `RiskFactorAdjustableModel` and I see they do use `adjust_risk_factors` so does it mean that all dynamic models will adjust the risk factors?

src/HealthGPS/risk_factor_adjustable_model.cpp
alexdewar: `dynamic_cast` lets you cast from a parent class to a child class, which you otherwise wouldn't be able to do. If it *isn't* an instance of your child class, it returns `nullptr`.

src/HealthGPS/risk_factor_adjustable_model.cpp
```diff
@@ -0,0 +1,170 @@
...
+        context.scenario().channel().send(std::make_unique<RiskFactorAdjustmentMessage>(
+            context.current_run(), context.time_now(), std::move(adjustments)));
+    }
+}
+
+RiskFactorSexAgeTable
+RiskFactorAdjustableModel::calculate_adjustments(RuntimeContext &context,
+                                                 const std::vector<core::Identifier> &keys) const {
+    auto age_range = context.age_range();
+    auto max_age = age_range.upper() + 1;
```
alexdewar, line 95: Why the `+ 1`?

src/HealthGPS/riskfactor.cpp
```diff
@@ -3,9 +3,8 @@
 namespace hgps {
 
 RiskFactorModule::RiskFactorModule(
-    std::map<RiskFactorModelType, std::unique_ptr<RiskFactorModel>> models,
-    const RiskfactorAdjustmentModel &adjustments)
-    : models_{std::move(models)}, adjustment_{adjustments} {
+    std::map<RiskFactorModelType, std::unique_ptr<RiskFactorModel>> models)
+    : models_{std::move(models)} {
```
alexdewar, line 7: I guess this map needs to be ordered?

alexdewar (approved):
I've made a few queries but it all looks good to me.

I should say though that I don't feel I have a particularly good handle on this bit of the code, so my review is a bit superficial. If you've tested it all though, then I think it's fine :smile:

On that note, this could use some tests, though I know you're probably too time-pressured to do this right now. Maybe open an issue?

src/HealthGPS.Console/csvparser.cpp
```diff
@@ -152,11 +152,11 @@ hgps::core::DataTable load_datatable_from_csv(const poco::FileInfo &file_info) {
     return out_table;
 }
 
-std::map<hgps::core::Identifier, std::vector<double>>
+std::unordered_map<hgps::core::Identifier, std::vector<double>>
```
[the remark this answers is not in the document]
jamesturner246: Maps/sets are implemented with balanced binary search trees, whereas unordered_maps/unordered_sets are implemented with hash mappings. Therefore, the ave access speed of maps/sets is on the order of O(log n), whereas, for the unordered variants, the ave access speed is O(1). Smaller containers benefit from the binary search on average, but large containers (like a potentially large risk factor set) benefit from the hashing.

src/HealthGPS/map2d.h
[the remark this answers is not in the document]
jamesturner246: Yes, exactly. HealthGPS passes these string-like things around called `Identifier`s. @alexdewar also brought this up about maybe preferring integer indexing instead of these things. Cost is slight readability reduction -- index `3` versus `"age"_id`.

src/HealthGPS/risk_factor_adjustable_model.h
jamesturner246: `age` is an integer type, coming directly from `Person.age`. In the case you highlight, the innermost vector element is sized 0-max_age, so we can index it with a plain integer.

src/HealthGPS/risk_factor_adjustable_model.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+/// @brief Defines the first statistical moment type
+struct FirstMoment {
+  public:
+    void append(double value) noexcept {
+        count_++;
+        sum_ += value;
+    }
+
+    double mean() const noexcept { return (count_ > 0) ? (sum_ / count_) : 0.0; }
```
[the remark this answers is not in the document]
jamesturner246: Throwing an exception would break, since there many not be a person of a particular sex and age in a population, thus `0.0 / 0`. NaN is preferable in this case. One could even just return `sum_ / count_`.

src/HealthGPS/riskfactor.cpp
jamesturner246: Actually no, they're explicitly indexed. I just missed it!

src/HealthGPS/risk_factor_adjustable_model.cpp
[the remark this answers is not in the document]
jamesturner246: ... On second thought, no. What if there *is not* a person of a given sex and age in the *baseline* scenario, but there *is* a person of that sex and age in the *intervention* scenario? That person's otherwise valid risk factor would have NaN added to it.

src/HealthGPS/risk_factor_adjustable_model.cpp
jamesturner246: Correct. Exactly that.

src/HealthGPS/risk_factor_adjustable_model.cpp
jamesturner246: I habitually pass things around in packaged form, but I see your point. I'll change the ones I spot.

src/HealthGPS/dynamic_hierarchical_linear_model.cpp
jamesturner246: Yes, your impression is correct. In future, once generic modules #231 is implemented, I would like to separate baseline adjustments out completely into it's own module, where a user would have one or more of these in their module ordered list (for timing), with their own variables they operate on. Time and existing code constraints mean I need to squash it in here for now though.

src/HealthGPS/risk_factor_adjustable_model.cpp
jamesturner246: I'm not completely sure I understand the question, but the pipe is unidirectional; baseline -> intervention. Does that answer?

jamesturner246:
From above:

> * Are static models expected to have to adjust the risk factors? If no, why do they inherit from `RiskFactorAdjustableModel`? If yes, where is that adjustment called?
 
This is just an artefact of having to squash many different functionalities into 'static' and 'dynamic' models. I think the whole static/dynamic model distinction has lost it's meaning and is generally unhelpful now, given that both model types have both initialisation and update methods. I think, from now, just ignore 'static' and 'dynamic', and focus more on the `generate_risk_factors` and `update_risk_factors`, and it might be easier to follow.

> * I see that dynamic models do inherit from `RiskFactorAdjustableModel` and I see they do use `adjust_risk_factors` so does it mean that all dynamic models will adjust the risk factors?

The reason one of the 'static' risk factor models (`StaticLinearModel`) inherits from `RiskFactorAdjustableModel` is because it eventually will need to use the baseline adjustment feature. For example, in the next step, when the food group risk factors are initialised. Hope that's clearer -- ask away if not.

jamesturner246:
Actionable items, for my convenience: 

* Passing explicit values in structures that are needed, instead of whole structure
* model_parser.cpp:227: deal with hard coded number (maybe leave it for an issue)
* HealthGPS.Test.cpp:409: `ASSERT_THROW(RiskFactorModule({}), std::invalid_argument);`
* riskfactor.cpp:6: use an `unordered_map`

src/HealthGPS/riskfactor.cpp
jamesturner246: Having said that, there's just two elements: static and dynamic. In this case, a plain ol' `std::map` would in fact be faster than a `std::unordered_map`t. In any case, this will eventually be replaced with an ordered container, like `std::vector` for #231.

src/HealthGPS.Console/model_parser.cpp
```diff
@@ -193,13 +219,12 @@ load_dynamic_risk_model_definition(const std::string &model_name, const poco::js
...
-load_ebhlm_risk_model_definition(const poco::json &opt) {
+load_ebhlm_risk_model_definition(const poco::json &opt, const host::Configuration &config) {
     MEASURE_FUNCTION();
-    auto percentage = 0.05;
-    std::map<hgps::core::Identifier, hgps::core::Identifier> variables;
-    std::map<hgps::core::IntegerInterval, hgps::AgeGroupGenderEquation> equations;
 
     auto info = poco::LiteHierarchicalModelInfo{};
+
+    auto percentage = 0.05;
```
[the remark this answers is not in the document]
jamesturner246: Opened issue #245.

src/HealthGPS/risk_factor_adjustable_model.cpp
jamesturner246: I've only addressed this instance, but I'll keep this in mind going forwards.

src/HealthGPS/risk_factor_adjustable_model.cpp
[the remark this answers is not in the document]
alexdewar: Ah ok. Maybe add a comment saying that then?

src/HealthGPS/riskfactor.cpp
alexdewar: Good point! Forget about it for now then

src/HealthGPS/risk_factor_adjustable_model.cpp
dalonsoa: No quite. My point is that this function does too many things depending on what the scenario is and, it seems to me, it also calls itself to go from baseline to intervention. I'm not sure how the messaging system work in C++, but it feels this function should be split into multiple ones doing one thing only each (and possibly calling each other).

dalonsoa: But not a barrier for this PR.

dalonsoa (approved):
The tests for linux-clang are failing, but this looks good, otherwise.

src/HealthGPS/risk_factor_adjustable_model.cpp
jamesturner246: Will make an issue and think about it.

src/HealthGPS/risk_factor_adjustable_model.cpp
jamesturner246: Renamed to `age_count`.

PR #248, opened 2023-11-07
Nutrient (food group) and Physical Activity models

### LINKED ISSUE #240: Nutrient model and Physical Activity model

jamesturner246:
Fixes #240

### DIFF
--- a/example_new/KevinHall.json
+++ b/example_new/KevinHall.json
@@ -67,43 +67,6 @@
             "Description": "string"
         }
     },
-    "RuralPrevalence": [
-        {
-            "Name": "Under18",
-            "Female": 0.65,
-            "Male": 0.65
-        },
-        {
-            "Name": "Over18",
-            "Female": 0.6,
-            "Male": 0.6
-        }
-    ],
-    "IncomeModels": [
-        {
-            "Name": "Category_1",
-            "Intercept": 0.0,
-            "Coefficients": {}
-        },
-        {
-            "Name": "Category_2",
-            "Intercept": -0.0023671151435328,
-            "Coefficients": {
-                "Gender": 0.0244784583947992,
-                "Over18": 0.276669591410528,
-                "Sector": -0.491794449118805
-            }
-        },
-        {
-            "Name": "Category_3",
-            "Intercept": 0.0187952471153142,
-            "Coefficients": {
-                "Gender": 0.02018185354194,
-                "Over18": 0.607099252829797,
-                "Sector": -1.5870837069653
-            }
-        }
-    ],
     "AgeMeanHeight": {
         "Male": [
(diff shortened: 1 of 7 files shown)


example_new/KevinHall.json
```diff
@@ -24,33 +24,28 @@
...
-            "Name": "Carbohydrate",
+            "Name": "FoodCarbohydrate",
             "Nutrients": {
                 "Carbohydrate": 1.0
             },
             "Price": null
         },
         {
-            "Name": "Fat",
+            "Name": "FoodFat",
```
alexdewar, line 34: Why are you prefixing things with food?

src/HealthGPS.Console/model_parser.cpp
```diff
@@ -146,64 +146,41 @@ load_staticlinear_risk_model_definition(const poco::json &opt, const host::Confi
...
-    }
-    for (size_t j = 0; j < quantiles_male.num_rows(); j++) {
-        weight_quantiles[hgps::core::Gender::male].push_back(
-            std::any_cast<double>(quantiles_male.column(0).value(j)));
-    }
+    // Risk factor names, models, parameters and correlation matrix.
+    std::vector<hgps::core::Identifier> names{};
+    std::vector<hgps::LinearModelParams> models{};
+    std::vector<double> lambda{};
+    std::vector<double> stddev{};
```
alexdewar, line 153: These will already be default-initialised, so you don't need the curlies:

```suggestion
    std::vector<hgps::core::Identifier> names;
    std::vector<hgps::LinearModelParams> models;
    std::vector<double> lambda;
    std::vector<double> stddev;
```

You only need those for basic types like ints which will otherwise be uninitialised.

alexdewar (approved):
A couple of queries, but otherwise LGTM.

example_new/KevinHall.json
jamesturner246: To avoid confusion with 'actual' nutrients in the risk factor mapping. These atre just dummy/meta foods, where the 'food' protein provides 100% 'real' protein, and nothing else.

jamesturner246: To avoid confusion with 'actual' nutrients in the risk factor mapping. These are just dummy/meta foods, where i.e. the 'food' protein provides 100% 'real' protein, and nothing else.

jamesturner246: To avoid confusion with 'actual' nutrients in the risk factor mapping. These are just dummy/meta foods, where i.e. the 'food' protein provides 100% 'real' protein, and nothing else.

src/HealthGPS.Console/model_parser.cpp
jamesturner246: Ah, okay. I will fix these in the next PR, if that's okay with you.

Issue #237, opened 2023-10-27
Fix clang-14 build (problem with chrono header)


jamesturner246:
Ubuntu clang-14 builder fails with 

```
 Error: /usr/bin/../lib/gcc/x86_64-linux-gnu/13/../../../../include/c++/13/chrono:2320:48: error: call to consteval function 'std::chrono::hh_mm_ss::_S_fractional_width' is not a constant expression
        static constexpr unsigned fractional_width = {_S_fractional_width()};
```

See https://github.com/actions/runner-images/issues/8659

I have disabled the buildder for now, but this needs fixing and/or a different/newer clang builder needs to be adopted.

jamesturner246:
I think it just needs replacing with a clang-15 or newer builder. So many problems with this specific version for C++20.

jamesturner246:
The Ubuntu builders we use now have a sufficient clang version, plus we are not currently testing on clang anyway.

Closing.

PR #250, opened 2023-11-08
Initialise weights

dalonsoa:
This PR adds an initial implementation for the initialisation of weights. The functionality to get the weight quantiles just returns 1 if quantiles are not available, which is the case until #251 is implemented. This means that there is no distribution of weights but just a single value, but should work well as a starting point.


dalonsoa:
Test are failing for clang, but no idea why since the error reported has nothing to do with what I've been doing... I think.

alexdewar:
@dalonsoa The `clang` runner has been failing for a while now... It isn't to do with anything you've changed.

@jamesturner246 Maybe we should update the branch protection rules to remove it as a requirement for PRs?

src/HealthGPS/static_linear_model.cpp
```diff
@@ -85,6 +98,33 @@ void StaticLinearModel::linear_approximation(Person &person) {
     }
 }
 
+/// Initialises the weight of a person.
+///
+/// It uses the baseline adjustment to get its initial value, based on its sex and age.
+/// @param person The person fo initialise the weight for.
+/// @generator Random number generator for the simulation.
+void StaticLinearModel::initialise_weight(Person &person, Random generator) {
```
alexdewar, line 106: Make `generator` a reference to avoid an unnecessary copy. If it can be const ref, that would be even better, but that might not be possible.

alexdewar (changes requested):
A couple of minor suggestions, but otherwise LGTM :smile:

src/HealthGPS/static_linear_model.cpp
```diff
@@ -85,6 +98,33 @@ void StaticLinearModel::linear_approximation(Person &person) {
...
+/// @gender The gender of the person.
+/// @generator Random number generator for the simulation.
+double StaticLinearModel::get_weight_quantile(core::Gender gender, Random generator) {
+
+    // TODO: Remove when fully implemented
+    if (weight_quantiles_.empty()) {
+        return 1.0;
+    }
+
+    int index = static_cast<int>(generator.next_double() * weight_quantiles_.at(gender).size());
```
jamesturner246, line 124: Furthermore, `Random` has a `next_int(int max_val)` method, so this can be:
```suggestion
    auto index = static_cast<size_t>(generator.next_int(weight_quantiles_.at(gender).size()));
```

src/HealthGPS/static_linear_model.cpp
```diff
@@ -85,6 +98,33 @@ void StaticLinearModel::linear_approximation(Person &person) {
...
+    auto weight_bl = get_risk_factor_expected().at(person.gender, "Weight"_id).at(person.age);
+    auto weight_quantile = get_weight_quantile(person.gender, generator);
+    person.risk_factors["Weight"_id] = weight_bl * weight_quantile;
+}
+
+/// Returns the weight quantile for the given gender.
+///
+/// @gender The gender of the person.
+/// @generator Random number generator for the simulation.
+double StaticLinearModel::get_weight_quantile(core::Gender gender, Random generator) {
```
jamesturner246, line 117: Same here: `generator` should be a reference, and `const` if possible.

jamesturner246 (changes requested):
Okay so far, and `adjust_risk_factors` is now in the correct place. A few minor things.

jamesturner246:
Hi @alexdewar. Already done. The clang test will always fail, due to #237, but it won't block the PR as it's not required. I haven't looked yet, but if there isn't a more modern clang builder, maybe we just disable it entirely.

alexdewar:
Okey doke. That makes sense.

We can probably get the latest clang from a PPA, possibly the same one we use for `gcc-latest`.

alexdewar (approved):
LGTM :smile:

jamesturner246 (approved):
LGTM

dalonsoa:
It seems the following line is now causing an error in Windows - and only in the CI, not in my computer:

https://github.com/imperialCHEPI/healthgps/blob/5a1da86fb15976e08a7a49a0472c9d2e7d1340a3/src/HealthGPS/static_linear_model.cpp#L124

Warning treated as error:
```output
D:\a\healthgps\healthgps\src\HealthGPS\static_linear_model.cpp(124): error C2220: the following warning is treated as an error
D:\a\healthgps\healthgps\src\HealthGPS\static_linear_model.cpp(124): warning C4267: 'argument': conversion from 'size_t' to 'int', possible loss of data
```

Do you have any suggestions?

jamesturner246:
Humm.. Interesting that the Windows compiler complains, of all of them. Generally speaking it's fine to index a vector with an integer, as long as it's not above the max `size_t` value divided by 2, which is unlikely here. Probably safe to just remove the `size_t` static cast then, if it keeps Windows happy.

dalonsoa:
If there's no explicit cast, there's an implicit one when when indexing the vector, and Windows MSVC seems to be equally un-happy about it. Is there any way to ignore that warning in that line, something equivalent to `type: ignore` in Python?

jamesturner246:
Fair enough. I'd say your original `next_double` solution is fine for now then.

PR #252, opened 2023-11-09
Read weight quantiles

### LINKED ISSUE #251: Read weight quantiles
Weight quantiles are required to initialise (and in the future update) the weights. So this issue relates to implementing the reading of such files and passing the relevant data to the `static_linear_model`.

dalonsoa:
Implements the logic to read the CSV files with the weight quantiles. The relevant files are added and the `StaitcLinear.json` file is modified to actually use that data. It also updates the static linear model and its definition to actually handle this information, removing the placeholders used before. 

Close #251


example_new/new_data/weight_quantiles_NCDRisk_female.csv
alexdewar, line 1: Where did this data come from? Is it from the researchers or is it made up?

src/HealthGPS.Console/model_parser.cpp
```diff
@@ -140,13 +140,31 @@ std::unique_ptr<hgps::StaticLinearModelDefinition>
...
+        {hgps::core::Gender::female, std::vector<double>(quantiles_female.num_rows())},
+        {hgps::core::Gender::male, std::vector<double>(quantiles_female.num_columns())}};
+    for (size_t j = 0; j < quantiles_female.num_rows(); j++) {
+        weight_quantiles[hgps::core::Gender::female][j] =
+            std::any_cast<double>(quantiles_female.column(0).value(j));
+    }
+    for (size_t j = 0; j < quantiles_male.num_rows(); j++) {
+        weight_quantiles[hgps::core::Gender::male][j] =
+            std::any_cast<double>(quantiles_male.column(0).value(j));
+    }
```
alexdewar, line 164: I guess you intended the male vector to be `quantiles_male.num_rows()` in size?

That aside, it's better not to resize your vectors upfront, but to resize them as you go along. If you want to hint to the compiler that you know the final size you can use the `reserve()` method.

I think this should work:

```suggestion
    std::map<hgps::core::Gender, std::vector<double>> weight_quantiles = {
        {hgps::core::Gender::female, {}},
        {hgps::core::Gender::male, {}};
    weight_quantiles[hgps::core::Gender::female].reserve(quantiles_female.num_rows());
    weight_quantiles[hgps::core::Gender::male].reserve(quantiles_male.num_rows());
    for (size_t j = 0; j < quantiles_female.num_rows(); j++) {
        weight_quantiles[hgps::core::Gender::female].push_back(std::any_cast<double>(quantiles_female.column(0).value(j)));
    }
    for (size_t j = 0; j < quantiles_female.num_rows(); j++) {
        weight_quantiles[hgps::core::Gender::male].push_back(std::any_cast<double>(quantiles_male.column(0).value(j)));
    }
```

That said, this is rather ugly. I see this is where you wanted to use #253.

The problem is we get a `DataTypeColumn` here (the abstract base class), whereas your fix applies to `PrimitiveDataTypeColumn` -- I can't think of a better way to reorganise this API without digging into it a bit more though (which I'm happy to do, @jamesturner246!).

One option would be to use `dynamic_cast` to convert the `DataTypeColumn`s into `PrimitiveDataTypeColumn<double>`s, then we can use your fix in #253 to get at the underlying `std::vector`s. What do you reckon, @jamesturner246?

src/HealthGPS/static_linear_model.cpp
```diff
@@ -116,21 +115,18 @@ void StaticLinearModel::initialise_weight(Person &person, Random &generator) {
 /// @generator Random number generator for the simulation.
 double StaticLinearModel::get_weight_quantile(core::Gender gender, Random &generator) {
 
-    // TODO: Remove when fully implemented
-    if (weight_quantiles_.empty()) {
-        return 1.0;
-    }
-
-    auto index = static_cast<size_t>(generator.next_int(weight_quantiles_.at(gender).size()));
+    int index = static_cast<int>(generator.next_double() * weight_quantiles_.at(gender).size());
```
alexdewar, line 118:
```suggestion
    auto index = static_cast<size_t>(generator.next_double() * weight_quantiles_.at(gender).size());
```

src/HealthGPS/static_linear_model.cpp
```diff
@@ -116,21 +115,18 @@ void StaticLinearModel::initialise_weight(Person &person, Random &generator) {
...
 StaticLinearModelDefinition::StaticLinearModelDefinition(
     RiskFactorSexAgeTable risk_factor_expected, std::vector<LinearModelParams> risk_factor_models,
-    Eigen::MatrixXd risk_factor_cholesky)
+    Eigen::MatrixXd risk_factor_cholesky,
+    const std::map<core::Gender, std::vector<double>> weight_quantiles)
     : RiskFactorAdjustableModelDefinition{std::move(risk_factor_expected)},
       risk_factor_models_{std::move(risk_factor_models)},
-      risk_factor_cholesky_{std::move(risk_factor_cholesky)} {
+      risk_factor_cholesky_{std::move(risk_factor_cholesky)},
+      weight_quantiles_{std::move(weight_quantiles)} {
```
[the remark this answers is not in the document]
alexdewar: I don't understand this complaint. `weight_quantiles` doesn't look `const` to me :raised_eyebrow:

src/HealthGPS.Console/model_parser.cpp
alexdewar: PS -- The way you've done it is also fine (aside from the sizing problem) and if we're going to rework the API anyway, there's not much point in mucking about with it now if it works.

src/HealthGPS/static_linear_model.cpp
```diff
@@ -21,10 +21,9 @@ StaticLinearModel::StaticLinearModel(
     if (!risk_factor_cholesky_.allFinite()) {
         throw core::HgpsException("Risk factor Cholesky matrix contains non-finite values");
     }
-    // TODO: Uncomment when fully implemented
-    // if (weight_quantiles_.empty()) {
-    //    throw core::HgpsException("Weigth quantiles dictionary is empty");
-    //}
+    if (weight_quantiles_.empty()) {
+        throw core::HgpsException("Weigth quantiles dictionary is empty");
```
alexdewar, line 25:
```suggestion
        throw core::HgpsException("Weight quantiles dictionary is empty");
```

alexdewar (commented):
I found one small mistake, but otherwise it's all good except for a few queries. I think we need to rework the `DataTypeTableColumn` API a bit, because it's not v nice to use atm.

example_new/new_data/weight_quantiles_NCDRisk_female.csv
dalonsoa: From Ali, as far as I know.

src/HealthGPS.Console/model_parser.cpp
dalonsoa: Yep, this is where I was thinking on using that patch. 

I thought on using `push_back`, but for some reason I thought it was going to be less efficient that just allocating the space upfront, given that I knew it in advance, and just populating the elements.

Thanks for pointing the male/female issue. I'l fix it asap.

dalonsoa: If not using something similar to #253, the only other option I can think of is implementing my own parser to read those files, but it is a pity given that there is already perfectly fine functionality to do that.

src/HealthGPS/static_linear_model.cpp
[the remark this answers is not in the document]
dalonsoa: I've no idea. I don't see any constants around...

src/HealthGPS.Console/model_parser.cpp
alexdewar: It's likely to be a really minor performance difference in this case, but the reason *not* to resize containers when creating them is because the contents then have to be default-initialised (in this case that just means they'll be zeroed). If you use `reserve()` instead, you still allocate memory but there's no initialisation required.

alexdewar: > If not using something similar to #253, the only other option I can think of is implementing my own parser to read those files, but it is a pity given that there is already perfectly fine functionality to do that.

I *think* we could rework the API so that you could get the values out by doing something like:

```cpp
std::vector<double> my_values = column.values<double>();
```

If `column` contains `int`s or something else, you would get an exception.

dalonsoa: Keeping aside mentioning the specific `type`, this was my intention, but it seems I did it in the wrong place.

alexdewar: You would need to update `PrimitiveDataColumnType` as you've done as well, but we also need to figure out how to change the base class to make use of your new `values()` method. Python makes this so much easier :laughing:

dalonsoa:
@jamesturner246 , this seems to work after the last change. So if we are moving forward with the merging, I'd suggest we first merge this PR and then #250 .

src/HealthGPS.Console/model_parser.cpp
jamesturner246: I think this can wait until afterwards. Lots of changes going on there at the moment.

src/HealthGPS/static_linear_model.cpp
```diff
@@ -21,10 +21,9 @@ StaticLinearModel::StaticLinearModel(
     if (!risk_factor_cholesky_.allFinite()) {
         throw core::HgpsException("Risk factor Cholesky matrix contains non-finite values");
     }
-    // TODO: Uncomment when fully implemented
-    // if (weight_quantiles_.empty()) {
-    //    throw core::HgpsException("Weigth quantiles dictionary is empty");
-    //}
+    if (weight_quantiles_.empty()) {
+        throw core::HgpsException("Weight quantiles dictionary is empty");
+    }
```
jamesturner246, line 26: I'd add the same check to `*ModelDefinition` below.

jamesturner246 (approved):
Looks good. I'd just add the same vector checking to `ModelDefinition` ctor as well, and good to go.

src/HealthGPS/static_linear_model.cpp
dalonsoa: Done!

PR #253, opened 2023-11-09
Add a method to get all the values of a DataTableColumn

dalonsoa:
To simplify some of the coding when reading data. Maybe this is just me being naive...

### DIFF
--- a/src/HealthGPS.Core/column_primitive.h
+++ b/src/HealthGPS.Core/column_primitive.h
@@ -113,6 +113,9 @@ template <typename TYPE> class PrimitiveDataTableColumn : public DataTableColumn
     /// @return An iterator to the end
     IteratorType end() const { return IteratorType(*this, size()); }
 
+    /// Returns the underlying column data as a vector.
+    std::vector<value_type> values() { return data_; }
+
   private:
     std::string name_;
     std::vector<TYPE> data_;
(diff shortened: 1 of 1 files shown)


src/HealthGPS.Core/column_primitive.h
```diff
@@ -113,6 +113,9 @@ template <typename TYPE> class PrimitiveDataTableColumn : public DataTableColumn
     /// @return An iterator to the end
     IteratorType end() const { return IteratorType(*this, size()); }
 
+    /// Returns the underlying column data as a vector.
+    std::vector<value_type> values() { return data_; }
```
alexdewar, line 117: This should return a const ref, otherwise it'll unconditionally copy the vector. If you return a const ref then the caller has the option of copying it, but can also just use it as a reference.

```suggestion
    const std::vector<value_type> &values() { return data_; }
```

alexdewar (commented):
Aside from the comment, I'm just wondering if you have a place in mind where you want to use this? It def makes sense if you want the values as a `std::vector`, but if you just want to iterate over them, you can already do this directly because it defines `begin()` and `end()` methods.

dalonsoa:
There are a few places where this might be useful, for a start, here https://github.com/imperialCHEPI/healthgps/blob/13a6912d3a38aa72cfe25d3ba265e3bebbbac536/src/HealthGPS.Console/model_parser.cpp#L157 and a few lines below.

I could just re-define the quantiles as columns instead of vectors, but it seems overkilling for just a single column.

jamesturner246 (approved):
Fine for these purposes. We can think about the best way to use a `DataTable` in future.

dalonsoa:
Closing this since, although the idea goes in the right way, it needs a more elaborate solution for it to work.

Issue #222, opened 2023-10-11
Consider making `context` in `generate_` and `update_risk_factors` const


jamesturner246:
Right now it is not, and we don't need to write to the `RuntimeContext` object, only the `Population` object contained therein.

Requires changing the RF model interface `HierarchicalLinearModel` (a bad and misleading name -- should think about changing this soon) and derived classes.

alexdewar:
I've just been having a look at this and while it *might* be possible, I think it'll be a fair amount of work (there are >50 uses of `context.population()`).

Whether this is worth the effort or not kind of depends on what your longer-term refactoring plans are. Does pulling the population part out of context make sense?

alexdewar:
I'm wondering whether it would make sense to split up `RuntimeContext` into two different data structures: one for the immutable data (model parameters) and the other for mutable data (e.g. the population, maybe even the RNG).

jamesturner246:
I'm about to change a whole lot of the context structure shortly, so happy to put this in the backlog for now.

PR #258, opened 2023-11-14
`RuntimeContext` tweaks

alexdewar:
I made a couple of small changes to `RuntimeContext` while working on #222. It turns out that #222 isn't currently doable, because there are models which modify `Person`s retrieved from the `RuntimeContext`, so `RuntimeContext` can't be `const`. I think the changes are worth having anyway though.

As @dalonsoa suggested, it would be better to break up the `RuntimeContext` object into separate parts, some of which could be passed in as non-`const` references. Is there an issue open for this?

### DIFF
--- a/src/HealthGPS/runtime_context.h
+++ b/src/HealthGPS/runtime_context.h
@@ -60,7 +60,7 @@ class RuntimeContext {
 
     /// @brief Gets a reference to the engine random number generator
     /// @return Random number generator
-    Random &random() noexcept;
+    Random &random() const noexcept;
 
     /// @brief Gets a read-only reference to the hierarchical risk factors mapping
     /// @return Risk factor mapping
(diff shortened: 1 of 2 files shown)


src/HealthGPS/runtime_context.h
```diff
@@ -99,7 +103,7 @@ class RuntimeContext {
     std::reference_wrapper<EventAggregator> event_bus_;
     std::reference_wrapper<SimulationDefinition> definition_;
     Population population_;
-    Random generator_;
+    mutable Random generator_;
```
jamesturner246, line 106: Why mutable here?

jamesturner246 (approved):
Fair enough. LGTM. Just the one question.

src/HealthGPS/runtime_context.h
alexdewar: Because that means we can return a non-`const` ref to `generator_` from a `const` method (`random()`). I think it makes sense -- the internal state of your random number generator is not something you actually need to keep `const`. This opens the door for us using `const RuntimeContext &` variables in future.

jamesturner246: Ah I see. Kinda like saying const method will keep everything constant, besides these few things marked mutable. Thanks,

dalonsoa (approved):
LGTM... although I don't fully grasps the implication of these changes.

Issue #181, opened 2023-07-20
EMB: chicken and egg scenario


jamesturner246:
A few places in `EnergyBalanceModel::simulate_person` compute **new** state variables from other **initial** state variables, where they should probably be computed from other **new** state. For example, `delta` (used in `EE`) is computed using `BW_0`, when it should probably be computed using `BW`:

```cpp
double delta = ((1.0 - beta_TEF) * PAL - 1.0) * RMR / BW_0;
//                                                    ^ should be BW?
```

Untangling this is non-trivial, since the new `BW` is currently indirectly a function of `delta`, but this just doesn't feel right.

Will require liaison with the modelling team to discuss intention and solutions.

jamesturner246:
Another in the update phase: should height be updated *before* weight?

jamesturner246:
Same in kevin hall initialisation: it goes weight, then, height, then kevin hall params (glycogen, ECF, etcetera)...

jackolney:
[Kevin Hall Framework.pdf](https://github.com/imperialCHEPI/healthgps/files/13962177/Kevin.Hall.Framework.pdf) – this might contain the answer?

PR #262, opened 2023-11-14
Update and adjust weight

### LINKED ISSUE #260: Compute weight update with adjustments

jamesturner246:
This PR consists of a majority rewrite of the Kevin Hall variable update step in `KevinHallModel`, to allow setting Kevin Hall risk factors, adjusting them with a new baseline adjustment code, and communicating those adjustments to the intervention scenario.

Now, starting with a call to the update step `update_risk_factors`  of this model, the following things happen:

1. Nutrient and energy intake are initialised for newborns, and updated for others.
2. Under 19: KevinHall does not apply, so initialise. Over 19: KevinHall applies, so run model.
3. Compute weight adjust, and compensation to other variables, and send adjustments to intervention.
4. Under 19: Reinitialise KevinHall variables with adjustment. Over 19: adjust KevinHall variables.
5. Send adjustments to intervention, if applicable.

All the existing baseline adjustment code was removed, and I and another tidy up.

Fixes #260

### DIFF
--- a/src/HealthGPS/kevin_hall_model.cpp
+++ b/src/HealthGPS/kevin_hall_model.cpp
@@ -91,51 +91,7 @@ void KevinHallModel::update_risk_factors(RuntimeContext &context) {
         }
     }
 
-    // TODO: Compute target body weight.
-    const float target_BW = 100.0;
-
-    // Trial run.
-    double mean_sim_body_weight = 0.0;
-    double mean_adjustment_coefficient = 0.0;
-    for (auto &person : context.population()) {
-        // Ignore if inactive.
-        if (!person.is_active()) {
-            continue;
-        }
-
-        // Simulate person and compute adjustment coefficient.
-        SimulatePersonState state = simulate_person(person, 0.0);
-        mean_sim_body_weight += state.BW;
-        mean_adjustment_coefficient += state.adjust;
-    }
-
-    // Compute model adjustment term.
-    const size_t population_size = context.population().current_active_size();
-    mean_sim_body_weight /= population_size;
-    mean_adjustment_coefficient /= population_size;
-    double shift = (target_BW - mean_sim_body_weight) / mean_adjustment_coefficient;
-
-    // Final run.
-    for (auto &person : context.population()) {
-        // Ignore if inactive.
-        if (!person.is_active()) {
-            continue;
-        }
-
-        // TODO: Simulate person and update risk factors.
-        simulate_person(person, shift);
-        // SimulatePersonState state = simulate_person(person, shift);
-        // person.risk_factors["Height"_id] = state.H;
-        // person.risk_factors["Weight"_id] = state.BW;
-        // person.risk_factors["PhysicalActivity"_id] = state.PAL;
-        // person.risk_factors["RestingMetabolicRate"_id] = state.RMR;
-        // person.risk_factors["BodyFat"_id] = state.F;
-        // person.risk_factors["LeanTissue"_id] = state.L;
-        // person.risk_factors["ExtracellularFluid"_id] = state.ECF;
-        // person.risk_factors["Glycogen"_id] = state.G;
-        // person.risk_factors["EnergyExpenditure"_id] = state.EE;
-        // person.risk_factors["EnergyIntake"_id] = state.EI;
-    }
+    // TODO: Weight adjustment.
 }
 
 void KevinHallModel::initialise_nutrient_intakes(Person &person) const {
@@ -186,9 +142,6 @@ void KevinHallModel::initialise_energy_intake(Person &person) const {
     // Start with previous = current.
     double energy_intake = person.risk_factors.at("EnergyIntake"_id);
     person.risk_factors["EnergyIntake_previous"_id] = energy_intake;
-
-    // Begin at steady state (EI = EE).
-    person.risk_factors["EnergyExpenditure"_id] = energy_intake;
 }
 
 void KevinHallModel::update_energy_intake(Person &person) const {
@@ -223,7 +176,7 @@ void KevinHallModel::initialise_kevin_hall_state(Person &person,
     }
 
     // Get already computed values.
-    double height = person.risk_factors.at("Height"_id);
+    double height = 0; // TODO: person.risk_factors.at("Height"_id);
     double weight = person.risk_factors.at("Weight"_id);
     double physical_activity = person.risk_factors.at("PhysicalActivity"_id);
     double energy_intake = person.risk_factors.at("EnergyIntake"_id);
@@ -263,52 +216,57 @@ void KevinHallModel::initialise_kevin_hall_state(Person &person,
     person.risk_factors["BodyFat"_id] = body_fat;
     person.risk_factors["LeanTissue"_id] = lean_tissue;
     person.risk_factors["Intercept_K"_id] = K;
+
+    // TODO: Compute and set new energy expenditure.
+    // TODO: SAME FOR ADJUSTMENT FOR kevin_hall_min_age PEOPLE
 }
 
-SimulatePersonState KevinHallModel::simulate_person(Person &person, double shift) const {
-    // Initial simulated person state.
-    const double H = person.get_risk_factor_value("Height"_id);
-    const double BW_0 = person.get_risk_factor_value("Weight"_id);
-    const double PAL_0 = person.get_risk_factor_value("PhysicalActivity"_id);
-    const double F_0 = person.get_risk_factor_value("BodyFat"_id);
-    const double L_0 = person.get_risk_factor_value("LeanTissue"_id);
-    const double ECF_0 = person.get_risk_factor_value("ExtracellularFluid"_id);
-    const double G_0 = person.get_risk_factor_value("Glycogen"_id);
-    const double EI_0 = person.get_risk_factor_value("EnergyIntake"_id);
-    const double CI_0 = person.get_risk_factor_value("Carbohydrate"_id);
+void KevinHallModel::simulate_person(Person &person) const {
 
-    // TODO: Compute physical activity level.
-    double PAL = PAL_0;
-
-    // TODO: some nutrient values need saving from previous year.
-
-    // Updated energy intake and carbohydrate intake.
-    // TODO: above, set EI_0 and CI_0 to risk_factors[*_previous].
-    double EI = person.get_risk_factor_value("EnergyIntake"_id);
-    double CI = person.get_risk_factor_value("Carbohydrate"_id);
-
-    // Compute glycogen and water.
-    double G = compute_G(CI, CI_0, G_0);
-    double W = compute_W(G);
-
-    // Compute extracellular fluid.
-    double ECF = compute_ECF(EI, EI_0, CI, CI_0, ECF_0);
+    // Get initial body weight.
+    double BW_0 = person.risk_factors.at("Weight"_id);
 
     // Compute energy cost per unit body weight.
+    double PAL = person.risk_factors.at("PhysicalActivity"_id);
+    double H = 0; // TODO: person.risk_factors.at("Height"_id);
     double delta = compute_delta(person.age, person.gender, PAL, BW_0, H);
 
+    // Get carbohydrate intake and energy intake.
+    double CI_0 = person.risk_factors.at("Carbohydrate_previous"_id);
+    double CI = person.risk_factors.at("Carbohydrate"_id);
+    double EI_0 = person.risk_factors.at("EnergyIntake_previous"_id);
+    double EI = person.risk_factors.at("EnergyIntake"_id);
+    double delta_EI = EI - EI_0;
+
     // Compute thermic effect of food.
-    double TEF = compute_TEF(EI, EI_0);
+    double TEF = beta_TEF * delta_EI;
 
     // Compute adaptive thermogenesis.
-    double AT = compute_AT(EI, EI_0);
+    double AT = beta_AT * delta_EI;
 
-    // TODO: Compute intercept value.
-    const double K = 0.0; // TODO: Intercept value.
+    // Compute glycogen and water.
+    double G_0 = person.risk_factors.at("Glycogen"_id);
+    double G = compute_G(CI, CI_0, G_0);
+    double W = 2.7 * G;
+
+    // Compute extracellular fluid.
+    double Na_0 = person.risk_factors.at("Sodium_previous"_id);
+    double Na = person.risk_factors.at("Sodium"_id);
+    double delta_Na = Na - Na_0;
+    double ECF_0 = person.risk_factors.at("ExtracellularFluid"_id);
+    double ECF = compute_ECF(delta_Na, CI, CI_0, ECF_0);
+
+    // Get initial body fat and lean tissue.
+    double F_0 = person.risk_factors.at("BodyFat"_id);
+    double L_0 = person.risk_factors.at("LeanTissue"_id);
+
+    // Get intercept value.
+    double K = person.risk_factors.at("Intercept_K"_id);
 
     // Energy partitioning.
-    const double C = 10.4 * rho_L / rho_F;
-    const double p = C / (C + F_0);
+    double C = 10.4 * rho_L / rho_F;
+    double p = C / (C + F_0);
+    double x = p * eta_L / rho_L + (1.0 - p) * eta_F / rho_F;
 
     // First equation ax + by = e.
     double a1 = p * rho_F;
@@ -318,13 +276,11 @@ SimulatePersonState KevinHallModel::simulate_person(Person &person, double shift
     // Second equation cx + dy = f.
     double a2 = gamma_F + delta;
     double b2 = gamma_L + delta;
-    double c2 = EI - shift - K - TEF - AT - delta * (G + W + ECF);
+    double c2 = EI - K - TEF - AT - delta * (G + W + ECF);
 
     // Compute body fat and lean tissue steady state.
     double steady_F = -(b1 * c2 - b2 * c1) / (a1 * b2 - a2 * b1);
     double steady_L = -(c1 * a2 - c2 * a1) / (a1 * b2 - a2 * b1);
-
-    double x = p * eta_L / rho_L + (1.0 - p) * eta_F / rho_F;
 
     // Compute time constant.
     double tau = rho_L * rho_F * (1.0 + x) /
@@ -334,27 +290,18 @@ SimulatePersonState KevinHallModel::simulate_person(Person &person, double shift
     double F = steady_F - (steady_F - F_0) * exp(-365.0 / tau);
     double L = steady_L - (steady_L - L_0) * exp(-365.0 / tau);
 
-    // Compute adjustment coefficient.
-    double adjust = -(a1 - b1) * (1.0 - exp(-365.0 / tau)) / (a1 * b2 - a2 * b1);
-
     // Compute body weight.
     double BW = F + L + G + W + ECF;
 
-    // Compute energy expenditure.
-    double EE =
-        (shift + K + gamma_F * F + gamma_L * L + delta * BW + TEF + AT + EI * x) / (1.0 + x);
+    // TODO: CHECK EQN WITH ALI: Compute energy expenditure (in adjustment step and init).
+    // double EE = (K + gamma_F * F + gamma_L * L + delta * BW + TEF + AT + EI * x) / (1.0 + x);
 
-    // New simulated person state.
-    return SimulatePersonState{.H = H,
-                               .BW = BW,
-                               .PAL = PAL,
-                               .F = F,
-                               .L = L,
-                               .ECF = ECF,
-                               .G = G,
-                               .EE = EE,
-                               .EI = EI,
-                               .adjust = adjust};
+    // Set new state.
+    person.risk_factors.at("Glycogen"_id) = G;
+    person.risk_factors.at("ExtracellularFluid"_id) = ECF;
+    person.risk_factors.at("BodyFat"_id) = F;
+    person.risk_factors.at("LeanTissue"_id) = L;
+    person.risk_factors.at("Weight"_id) = BW;
 }
 
 double KevinHallModel::compute_G(double CI, double CI_0, double G_0) const {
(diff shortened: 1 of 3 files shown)


src/HealthGPS/kevin_hall_model.cpp
```diff
@@ -81,74 +94,114 @@ void KevinHallModel::update_risk_factors(RuntimeContext &context) {
...
     }
 
-    // TODO: Compute target body weight.
-    const float target_BW = 100.0;
-
-    // Trial run.
-    double mean_sim_body_weight = 0.0;
-    double mean_adjustment_coefficient = 0.0;
+    // Update: different model for children and adults.
     for (auto &person : context.population()) {
```
dalonsoa, line 104: I don't see any reason why this loop over population could not be merged with the previous one, saving one whole loop and checking if the person is active.

src/HealthGPS/kevin_hall_model.cpp
```diff
@@ -81,74 +94,114 @@ void KevinHallModel::update_risk_factors(RuntimeContext &context) {
...
+}
+
 void KevinHallModel::initialise_nutrient_intakes(Person &person) const {
 
     // Initialise nutrient intakes.
     set_nutrient_intakes(person);
 
-    // TODO: set old value to new value (only some nutrients)
+    // Start with previous = current.
+    double carbohydrate = person.risk_factors.at("Carbohydrate"_id);
```
dalonsoa, line 189: This is just a style question: why sometimes do you use `auto` and in others the explicit type?

src/HealthGPS/kevin_hall_model.cpp
```diff
@@ -168,18 +221,23 @@ void KevinHallModel::set_nutrient_intakes(Person &person) const {
...
+    person.risk_factors["EnergyIntake_previous"_id] = energy_intake;
 }
 
 void KevinHallModel::update_energy_intake(Person &person) const {
-    set_energy_intake(person);
 
-    // TODO: set old value to previous value
+    // Set previous energy intake.
+    double previous_energy_intake = person.risk_factors.at("EnergyIntake"_id);
+    person.risk_factors.at("EnergyIntake_previous"_id) = previous_energy_intake;
```
dalonsoa, line 237: Why do you need to save the previous value? (also in the previous function)

src/HealthGPS/kevin_hall_model.cpp
```diff
@@ -195,53 +253,110 @@ void KevinHallModel::set_energy_intake(Person &person) const {
...
+    double EI = person.risk_factors.at("EnergyIntake"_id);
+
+    // Body fat.
+    double F;
+    if (person.gender == core::Gender::female) {
+        F = BW * (0.14 * person.age + 39.96 * log(BW / pow(H / 100, 2.0)) - 102.01) / 100.0;
+    } else if (person.gender == core::Gender::male) {
+        F = BW * (0.14 * person.age + 37.31 * log(BW / pow(H / 100, 2.0)) - 103.94) / 100.0;
+    } else {
+        throw core::HgpsException("Unknown gender");
```
dalonsoa, line 277: Is it possible to get to this point of the simulation with an unknown gender?

src/HealthGPS/kevin_hall_model.cpp
```diff
@@ -195,53 +253,110 @@ void KevinHallModel::set_energy_intake(Person &person) const {
...
+    double BW = person.risk_factors.at("Weight"_id);
+    double PAL = person.risk_factors.at("PhysicalActivity"_id);
+    double EI = person.risk_factors.at("EnergyIntake"_id);
+
+    // Body fat.
+    double F;
+    if (person.gender == core::Gender::female) {
+        F = BW * (0.14 * person.age + 39.96 * log(BW / pow(H / 100, 2.0)) - 102.01) / 100.0;
+    } else if (person.gender == core::Gender::male) {
+        F = BW * (0.14 * person.age + 37.31 * log(BW / pow(H / 100, 2.0)) - 103.94) / 100.0;
```
dalonsoa, line 275: Personally, I'd put these two formulas in their own functions using simple double arguments, making them easy to test, and describe in docstring where the formulas are coming from (paper, reference, etc.)

src/HealthGPS/kevin_hall_model.cpp
```diff
@@ -195,53 +253,110 @@ void KevinHallModel::set_energy_intake(Person &person) const {
...
-    double G = compute_G(CI, CI_0, G_0);
-    double W = compute_W(G);
+    // Model intercept value.
+    double delta = compute_delta(person.age, person.gender, PAL, BW, H);
+    double K = EI - (gamma_F * F + gamma_L * L + delta * BW);
 
-    // Compute extracellular fluid.
-    double ECF = compute_ECF(EI, EI_0, CI, CI_0, ECF_0);
+    // Compute energy expenditure.
+    double C = 10.4 * rho_L / rho_F;
```
dalonsoa, line 298: I cannot find where these `rho` values are coming from... Where are they calculated or obtained?

src/HealthGPS/kevin_hall_model.cpp
```diff
@@ -195,53 +253,110 @@ void KevinHallModel::set_energy_intake(Person &person) const {
...
+    double p = C / (C + F);
+    double x = p * eta_L / rho_L + (1.0 - p) * eta_F / rho_F;
+    double EE = compute_EE(BW, F, L, EI, K, delta, x);
```
dalonsoa, line 301: All of these formulas and calculation looks like dark magic coming out of the blue. Maybe as a separate use given the time constrains, but I believe they should be separated in their own functions with proper explanations of their origin.

src/HealthGPS/kevin_hall_model.cpp
```diff
@@ -195,53 +253,110 @@ void KevinHallModel::set_energy_intake(Person &person) const {
...
+    person.risk_factors["LeanTissue"_id] = L;
+    person.risk_factors["Glycogen"_id] = G;
+    person.risk_factors["ExtracellularFluid"_id] = ECF;
+    person.risk_factors["Intercept_K"_id] = K;
+
+    // Set new energy expenditure (may not exist yet).
+    person.risk_factors["EnergyExpenditure"_id] = EE;
+}
+
+void KevinHallModel::kevin_hall_run(Person &person) const {
```
dalonsoa, line 314: Same here - not your fault! All of the formulas and calculations here are somewhat meaningless without an explanation, even dangerous because there's no way of checking their correctness.

dalonsoa (approved):
It looks Bradly good to me but, in general, there is a total lack of explanation of where all the formulas and mathematical calculations are coming from. Clearly they are representing a mathematical model - let me guess, the Kevin Hall model - which I guess is (or will be) explained in the documentation. But having such disconnection between code and explanation is dangerous and render the code an arbitrary collection of numbers and operations.

As I point out in one of the comments, I'll make a priority - for Ali - to write down in English, not just in maths, an explanation of these calculations, so they can be added to the relevant docstrings.

src/HealthGPS/kevin_hall_model.cpp
jamesturner246: Sadly it's  to do with a very strict and fragile ordering of variables... See the 'chicken and egg' issue #181. My brain cries tears at the thought of unwinding this right now, but I agree it should eventually be done.

src/HealthGPS/kevin_hall_model.cpp
jamesturner246: Good question.. I think I try to use the type when I want to make it obvious what it is, e.g. double, but use auto where it's a bit long type, like `map<int, map<Identifier, double>>`, or where it's obvious from the context (like `static_cast<int>(...)` or `get_complex()` etcetera. I think clang-tidy even enforces auto use with casting. I'm not sure if there are any useful standard guidelines though.

src/HealthGPS/kevin_hall_model.cpp
jamesturner246: This is to do with the computing the change in physical activity thing I mentioned in the previous PR: `double delta_EI = EI - EI_0`. Some parts of the Kevin Hall equations depend on previous risk factor values as well as current values.

src/HealthGPS/kevin_hall_model.cpp
jamesturner246: In theory, no, but I left it in to be safe. What do you do with code paths you expect not to be reached yourself?

src/HealthGPS/kevin_hall_model.cpp
jamesturner246: These are stored as private static variables in the `KevinHallModel` class -- see the header: `kevin_hall_model.h`.

src/HealthGPS/kevin_hall_model.cpp
jamesturner246: Agreed.

src/HealthGPS/kevin_hall_model.cpp
jamesturner246: Actually this was because I didn't want big ugly less readable conditional branches in the for loop. I probably could have squashed it together. in any case, the compiler is generally smart enough to reorganise it more efficiently, so it just boils down to readability preference in the end. Alex likes less indentation, for example.

src/HealthGPS/kevin_hall_model.cpp
dalonsoa: No, I think it is good practice to handle all possible cases, specially in Python. However, in here, the compiler will not let you assign `person.gender` anything other than `core::Gender::male` or `core::Gender::female`, so I'm not sure it this is even a possible case.

src/HealthGPS/kevin_hall_model.cpp
alexdewar: You'd *hope* the compiler would do something clever with it, but sadly I'm not sure you can rely on it. This is a slow path anyway though (right?) so performance doesn't really matter.

I think it would actually be more readable to squash the two loops together though as @dalonsoa suggests.

src/HealthGPS/kevin_hall_model.cpp
```diff
@@ -81,74 +94,114 @@ void KevinHallModel::update_risk_factors(RuntimeContext &context) {
...
-
-    // Trial run.
-    double mean_sim_body_weight = 0.0;
-    double mean_adjustment_coefficient = 0.0;
+    // Update: different model for children and adults.
     for (auto &person : context.population()) {
         // Ignore if inactive.
         if (!person.is_active()) {
             continue;
         }
```
alexdewar, line 108: Given that iterating over only the "active" people seems to be such a common operation, would it be useful to have some kind of helper function?

Obvs shouldn't be done as part of this PR

src/HealthGPS/kevin_hall_model.cpp
```diff
@@ -81,74 +94,114 @@ void KevinHallModel::update_risk_factors(RuntimeContext &context) {
...
+        auto &basePtr = message.value();
+        auto *messagePrt = dynamic_cast<KevinHallAdjustmentMessage *>(basePtr.get());
+        if (!messagePrt) {
+            throw core::HgpsException(
+                "Simulation out of sync, failed to receive a Kevin Hall adjustments message");
+        }
 
-    // Final run.
+        adjustments = messagePrt->data();
+    }
```
alexdewar, line 140: I'd make most of this block into a little helper function, returning a `std::optional<KevinHallAdjustmentMessage>`, then you can write:

```cpp
    // Intervention scenario: recieve adjustments from baseline scenario.
    else if (const auto &message = try_get_adjustment_message(context))
        adjustments = message->data();
    else
        throw core::HgpsException(
                "Simulation out of sync, failed to receive a Kevin Hall adjustments message");
```

src/HealthGPS/kevin_hall_model.cpp
```diff
@@ -81,74 +94,114 @@ void KevinHallModel::update_risk_factors(RuntimeContext &context) {
...
 
+KevinHallAdjustmentTable
+KevinHallModel::compute_kevin_hall_adjustments(Population &population) const {
+    auto expected = get_risk_factor_expected();
+    auto weight_means = compute_mean_weight(population);
+
+    // Compute adjustments.
+    auto adjustments = KevinHallAdjustmentTable{};
+    for (const auto &[sex, weight_means_by_sex] : weight_means) {
+        adjustments.emplace_row(sex, std::unordered_map<int, double>{});
```
alexdewar, line 173: I think this will work:

```suggestion
        adjustments.emplace_row(sex, {});
```

src/HealthGPS/kevin_hall_model.cpp
alexdewar: You can even make `clang-tidy` enforce use of `auto` *everywhere* if you're particularly keen on it :laughing:. I agree with James though that it's usually nice to be explicit for short type names though.

src/HealthGPS/kevin_hall_model.cpp
```diff
@@ -168,18 +221,23 @@ void KevinHallModel::set_nutrient_intakes(Person &person) const {
...
-    // TODO: set old value to previous value
+    // Set previous energy intake.
+    double previous_energy_intake = person.risk_factors.at("EnergyIntake"_id);
+    person.risk_factors.at("EnergyIntake_previous"_id) = previous_energy_intake;
+
+    // Update energy intake.
+    set_energy_intake(person);
 }
 
 void KevinHallModel::set_energy_intake(Person &person) const {
```
alexdewar, line 243: Pedantic point, but I feel like the `set_` prefix makes it sound like more of a "setter" (i.e. for setting the intake to a specified number). How about `calculate_`/`recalculate_`?

src/HealthGPS/kevin_hall_model.cpp
alexdewar: If someone was being really perverse I *think* they could do something like `static_cast<core::Gender>(100)` to get a non-male, non-female value, but I don't think we should worry about that case. It's more of a concern when you are casting a user-supplied number to an `enum`, but that doesn't apply here.

src/HealthGPS/kevin_hall_model.cpp
```diff
@@ -195,53 +253,110 @@ void KevinHallModel::set_energy_intake(Person &person) const {
...
+    } else if (person.gender == core::Gender::male) {
+        F = BW * (0.14 * person.age + 37.31 * log(BW / pow(H / 100, 2.0)) - 103.94) / 100.0;
+    } else {
+        throw core::HgpsException("Unknown gender");
+    }
 
-    // TODO: Compute physical activity level.
-    double PAL = PAL_0;
+    if (F < 0.0) {
+        // HACK: deal with negative body fat.
```
alexdewar, line 281: What would give us negative body fat?

src/HealthGPS/kevin_hall_model.cpp
alexdewar: In an ideal world, they would also help us write meaningful tests for them.

src/HealthGPS/kevin_hall_model.cpp
```diff
@@ -267,66 +380,47 @@ SimulatePersonState KevinHallModel::simulate_person(Person &person, double shift
...
 
-double KevinHallModel::compute_delta(double PAL, double BW, double H, unsigned int age,
-                                     core::Gender gender) const {
+double KevinHallModel::compute_delta(int age, core::Gender sex, double PAL, double BW,
+                                     double H) const {
     // Resting metabolic rate (Mifflin-St Jeor).
-    double RMR = 9.99 * BW + 6.25 * H * 100.0 - 4.92 * age;
-    RMR += gender == core::Gender::male ? 5.0 : -161.0;
+    double RMR = 9.99 * BW + 6.25 * H - 4.92 * age;
+    RMR += sex == core::Gender::male ? 5.0 : -161.0;
```
alexdewar, line 407: I feel like these gender-specific values should all be bundled into a `struct` or something.

```cpp
struct FemaleConsts {
    static double my_value = 1.0;
    // etc.
};

struct MaleConsts {
    static double my_value = 2.0;
    // etc.
};
```

You could even define a helper macro for this sort of situation:

```cpp
#define SEX_VALUE(sex, name) \
((sex == core::Gender::male) ? MaleConsts::name : FemaleConsts::name)
```

src/HealthGPS/kevin_hall_model.cpp
```diff
@@ -361,6 +504,69 @@ double KevinHallModel::get_weight_quantile(double epa_quantile, core::Gender sex
...
+
+      private:
+        double sum_{};
+        int count_{};
+    };
+
+    // Compute sums and counts of weight powers for sex and age.
+    auto sumcounts = UnorderedMap2d<core::Gender, int, SumCount>{};
+    sumcounts.emplace_row(core::Gender::female, std::unordered_map<int, SumCount>{});
+    sumcounts.emplace_row(core::Gender::male, std::unordered_map<int, SumCount>{});
```
alexdewar, line 528:
```suggestion
    sumcounts.emplace_row(core::Gender::female, {});
    sumcounts.emplace_row(core::Gender::male, {});
```

src/HealthGPS/kevin_hall_model.cpp
```diff
@@ -361,6 +504,69 @@ double KevinHallModel::get_weight_quantile(double epa_quantile, core::Gender sex
...
+    for (const auto &person : population) {
+        if (!person.is_active()) {
+            continue;
+        }
+
+        double value;
+        if (power.has_value()) {
+            value = pow(person.risk_factors.at("Weight"_id), power.value());
```
alexdewar, line 536: Out of interest, can powers be floating-point numbers or are they always integers?

src/HealthGPS/kevin_hall_model.cpp
```diff
@@ -361,6 +504,69 @@ double KevinHallModel::get_weight_quantile(double epa_quantile, core::Gender sex
...
+        } else {
+            value = person.risk_factors.at("Weight"_id);
+        }
+
+        sumcounts.at(person.gender)[person.age].append(value);
+    }
```
alexdewar, line 542: I like the way you've implemented this, but this code is a perfect example of how things would be easier if HGPS used a `struct`-of-arrays design cf. an array-of-`struct`s one. Imagine if you could get men's and women's weights as two vectors instead. It would make life so much easier!

alexdewar (approved):
Some minor suggestions and queries, but nothing that should be a blocker. Otherwise LGTM!

src/HealthGPS/kevin_hall_model.cpp
jamesturner246: This unfortunately changed in a later PR. The ordering of this model is very brittle (#181), especially now with the latest draft PR. I'd prefer to address this then.

src/HealthGPS/kevin_hall_model.cpp
jamesturner246: There are other functions dotted around which construct new, or reinitialise or otherwise change `Person`s. As you say, not worth the risk.

src/HealthGPS/kevin_hall_model.cpp
jamesturner246: Agree. Less bug prone. I'll open an issue.

jamesturner246: #267.

src/HealthGPS/kevin_hall_model.cpp
jamesturner246: Strictly they're the the type of the base and return. Either `float` or `double`.

src/HealthGPS/kevin_hall_model.cpp
jamesturner246: Agree. Soon I hope.

src/HealthGPS/kevin_hall_model.cpp
jamesturner246: That's what these `set_` functions are essentially doing: setting risk factors in people. To me, `calculate_` implies returning that value to caller.

src/HealthGPS/kevin_hall_model.cpp
jamesturner246: Who knows? I don't. :shrug:

src/HealthGPS/kevin_hall_model.cpp
jamesturner246: Problem is I have no idea what these are. I think this should be part of a larger reorganisation and labelling effort down the line, when we have time to get them to go through it and name everything.

jamesturner246: #268

src/HealthGPS/kevin_hall_model.cpp
```diff
@@ -67,6 +75,12 @@ void KevinHallModel::generate_risk_factors(RuntimeContext &context) {
...
     adjust_risk_factors(context, {"Weight"_id});
+
+    // Initialise everyone.
+    for (auto &person : context.population()) {
+        initialise_kevin_hall_state(person);
+        compute_bmi(person);
+    }
 }
 
 void KevinHallModel::update_risk_factors(RuntimeContext &context) {
```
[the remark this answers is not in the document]
jamesturner246: I know, believe me.... -_-

src/HealthGPS/kevin_hall_model.cpp
jamesturner246: No good I'm afraid. Same reason.

src/HealthGPS/kevin_hall_model.cpp
jamesturner246: #267

src/HealthGPS/kevin_hall_model.cpp
jamesturner246: I'll leave this for #270, due to time constraints.

src/HealthGPS/kevin_hall_model.cpp
jamesturner246: Although maybe it's worth it for consistency. I'll change them in the next PR.

src/HealthGPS/kevin_hall_model.cpp
alexdewar: I meant for *this* code not in general. Are you using fractional powers or just integers? Because if it's just integers, this bit could be replaced with a `for`-loop.

src/HealthGPS/kevin_hall_model.cpp
alexdewar: Up to you. I agree that both naming schemes have problems :shrug:

src/HealthGPS/kevin_hall_model.cpp
jamesturner246: For part of the model, the height initialise and update I think, a real-valued `slope` parameter is used.

PR #264, opened 2023-11-15
Add computation of the BMI

### LINKED ISSUE #263: Compute BMI
NB: I would branch off either `height` or better yet `kevinhall` branches, as a lot has been added there.

This one should be fairly straightforward -- just implement a `compute_BMI` method to `KevinHall` which computes BMI from weight and height according to `BMI = W / H * H`.  There's a couple of stub methods in comments ready to go.

You'd need to access weight and height with `person.risk_factors.at("Height"_id)`, and save BMI with `person.risk_factors["BMI"_id]` (since there's no guarantee it exists in there yet.

This method should be called at the end of both the `generate_risk_factors` and `update_risk_factors`.  Also, if you like, you can remove the commented BMI hack in `person.cpp`, near the top.

Thanks! :)

dalonsoa:
Merges into Kevinhall branch.

Close #263

### DIFF
--- a/src/HealthGPS/kevin_hall_model.cpp
+++ b/src/HealthGPS/kevin_hall_model.cpp
@@ -381,7 +381,7 @@ double KevinHallModel::compute_EE(double BW, double F, double L, double EI, doub
 
 /// Compute's a person BMI assuming height in cm
 /// @param person The person to calculate the BMI for.
-void KevinHallModel::compute_bmi(Person &person) {
+void KevinHallModel::compute_bmi(Person &person) const {
     auto w = person.risk_factors.at("Weight"_id);
     auto h = person.risk_factors.at("Height"_id) * 100;
     person.risk_factors["BMI"_id] = w / (h * h);
--- a/src/HealthGPS/kevin_hall_model.cpp
+++ b/src/HealthGPS/kevin_hall_model.cpp
@@ -383,7 +383,7 @@ double KevinHallModel::compute_EE(double BW, double F, double L, double EI, doub
 /// @param person The person to calculate the BMI for.
 void KevinHallModel::compute_bmi(Person &person) const {
     auto w = person.risk_factors.at("Weight"_id);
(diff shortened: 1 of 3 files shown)


src/HealthGPS/kevin_hall_model.cpp
```diff
@@ -368,6 +379,14 @@ double KevinHallModel::compute_EE(double BW, double F, double L, double EI, doub
            (1 + x);
 }
 
+/// Compute's a person BMI assuming height in cm
+/// @param person The person to calculate the BMI for.
+void KevinHallModel::compute_bmi(Person &person) const {
+    auto w = person.risk_factors.at("Weight"_id);
+    auto h = person.risk_factors.at("Height"_id) * 100;
```
jamesturner246, line 386:
```suggestion
    auto h = person.risk_factors.at("Height"_id) / 100;
```
Since cm to metres

jamesturner246 (changes requested):
Thanks! Just the one errand multiplication, and LGTM.

src/HealthGPS/kevin_hall_model.cpp
```diff
@@ -368,6 +379,14 @@ double KevinHallModel::compute_EE(double BW, double F, double L, double EI, doub
            (1 + x);
 }
 
+/// Compute's a person BMI assuming height in cm
+/// @param person The person to calculate the BMI for.
+void KevinHallModel::compute_bmi(Person &person) {
```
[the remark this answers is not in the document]
jamesturner246: Ah yes, it's kinda confusing and arbitrary which keywords go in header vs source, but `const` is one that goes in both.

src/HealthGPS/kevin_hall_model.cpp
jamesturner246: Indeed, always the small things..

jamesturner246 (approved):
LGTM!

jamesturner246:
Thanks. Don't worry about the conflict. My bad, I'll merge it.

PR #265, opened 2023-11-15
Height init update

### LINKED ISSUE #107: Height variable

jamesturner246:
This PR adds the `initialise_height` and `update_height` methods, with supporting code, to the `KevinHallModel`.

A further reshuffling of `update_risk_factors` was required. It is ugly as hell, but the fragility of this horrible thing to update orderings makes it so. As a general rule to help follow, I have split the code mostly into newborn initialisation in the first half, and updating teenagers/adults in the second half. All three have different treatment.

Fixes #107 (finally)


jamesturner246:
I don't understand this failure on `linux, ubuntu-22.04, gcc-default, true`...

dalonsoa (changes requested):
You're right it looks pretty horrible. However, keeping aside major refactoring of the architecture of the code, I think it should be possible to "beautify" the code a bit by broken the `generate_risk_factors` method into smaller functions that take care of the individual use cases, that can be tested and explained appropriately (when the time comes). You might not have that much time now, but it might be necessary to keep the technical debt at bay, at least a bit.

alexdewar (approved):
A couple of comments, but otherwise LGTM.

src/HealthGPS/kevin_hall_model.cpp
```diff
@@ -95,16 +100,48 @@ void KevinHallModel::update_risk_factors(RuntimeContext &context) {
...
-        if (!person.is_active()) {
+        // Ignore if inactive or not a newborn.
+        if (!person.is_active() || person.age != 0) {
+            continue;
+        }
+
+        double W_expected = expected.at(person.gender, "Weight"_id).at(0);
+        double W_mean = W_newborn_means.at(person.gender, 0);
+        double adjustment = W_expected - W_mean;
+        person.risk_factors.at("Weight"_id) -= adjustment;
```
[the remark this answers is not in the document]
jamesturner246: You're right, `adjustment` should be added not subtracted, though not flipped. If expected is higher, we want a positive adjustment, so `expected - mean`, and then ADD this to the risk factor. I'll change it.

src/HealthGPS/kevin_hall_model.cpp
```diff
@@ -546,7 +594,8 @@ KevinHallAdjustmentTable KevinHallModel::compute_mean_weight(Population &populat
     sumcounts.emplace_row(core::Gender::female, std::unordered_map<int, SumCount>{});
     sumcounts.emplace_row(core::Gender::male, std::unordered_map<int, SumCount>{});
     for (const auto &person : population) {
-        if (!person.is_active()) {
+        bool age_mismatch = age.has_value() && person.age != age.value();
+        if (!person.is_active() || age_mismatch) {
             continue;
         }
```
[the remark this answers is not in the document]
jamesturner246: Done.

jamesturner246:
Okay. Phew. I've had a crack at refactoring `update_risk_factors`: Separate functions for newborn and non-newborn, separate functions for computing adjustments, removed some duplication, separate functions for sending the adjustments to intervention, and a bit more tidying. It's still not amazing, but at least a bit easier to follow.

See the `// TODO: FOR REFACTORING:` note: a big chunk of code (the entire `initialise_newborn` can be totally removed and replaced with a general initialisation function once #266 is implemented, making it look way more tidy. 

Also kinda fixes #269.

src/HealthGPS/kevin_hall_model.cpp
```diff
@@ -73,86 +73,156 @@ void KevinHallModel::generate_risk_factors(RuntimeContext &context) {
...
+    // Initialise nutrient and energy intake and weight for newborns.
     for (auto &person : context.population()) {
-        // Ignore if inactive.
+        // Ignore if inactive or not newborn.
         if (!person.is_active()) {
             continue;
         }
+        if (person.age != 0) {
+            continue;
+        }
```
dalonsoa, line 120: Just to visually reduce the amount of lines of code, yet keeping things simple:
```suggestion
        if (!person.is_active() || (person.age != 0)) {
            continue;
        }
```

src/HealthGPS/kevin_hall_model.cpp
```diff
@@ -73,86 +73,156 @@ void KevinHallModel::generate_risk_factors(RuntimeContext &context) {
...
+    // Adjust newborn weight to match expected.
+    auto adjustments = compute_weight_adjustments(context.population(), 0);
+    for (auto &person : context.population()) {
+        // Ignore if inactive or not newborn.
+        if (!person.is_active()) {
+            continue;
         }
+        if (person.age != 0) {
+            continue;
+        }
```
dalonsoa, line 148: As above:
```suggestion
        if (!person.is_active() || (person.age != 0)) {
            continue;
        }
```

src/HealthGPS/kevin_hall_model.cpp
```diff
@@ -73,86 +73,156 @@ void KevinHallModel::generate_risk_factors(RuntimeContext &context) {
...
+
+    // Initialise height and other Kevin Hall state for newborns.
+    for (auto &person : context.population()) {
+        // Ignore if inactive or not newborn.
+        if (!person.is_active()) {
+            continue;
+        }
+        if (person.age != 0) {
+            continue;
+        }
```
dalonsoa, line 167: As above:
```suggestion
        if (!person.is_active() || (person.age != 0)) {
            continue;
        }
```

src/HealthGPS/kevin_hall_model.cpp
```diff
@@ -73,86 +73,156 @@ void KevinHallModel::generate_risk_factors(RuntimeContext &context) {
...
+
+    // Update nutrient and energy intake for non-newborns.
+    for (auto &person : context.population()) {
+        // Ignore if inactive or newborn.
+        if (!person.is_active()) {
+            continue;
         }
+        if (person.age == 0) {
+            continue;
+        }
```
dalonsoa, line 185: Guess what? 😄 
```suggestion
        if (!person.is_active() || (person.age == 0)) {
            continue;
        }
```

src/HealthGPS/kevin_hall_model.cpp
```diff
@@ -73,86 +73,156 @@ void KevinHallModel::generate_risk_factors(RuntimeContext &context) {
...
-                "Simulation out of sync, failed to receive a Kevin Hall adjustments message");
+    // Update weight for non-newborns.
+    for (auto &person : context.population()) {
+        // Ignore if inactive or newborn.
+        if (!person.is_active()) {
+            continue;
+        }
+        if (person.age == 0) {
+            continue;
         }
```
dalonsoa, line 199:
```suggestion
        if (!person.is_active() || (person.age == 0)) {
            continue;
        }
```

src/HealthGPS/kevin_hall_model.cpp
```diff
@@ -73,86 +73,156 @@ void KevinHallModel::generate_risk_factors(RuntimeContext &context) {
...
+    // Adjust weight and other Kevin Hall state for non-newborns.
     for (auto &person : context.population()) {
-        // Ignore if inactive.
+        // Ignore if inactive or newborn.
         if (!person.is_active()) {
             continue;
         }
+        if (person.age == 0) {
+            continue;
+        }
```
dalonsoa, line 219:
```suggestion
        if (!person.is_active() || (person.age == 0)) {
            continue;
        }
```

src/HealthGPS/kevin_hall_model.cpp
```diff
@@ -162,29 +232,66 @@ void KevinHallModel::update_risk_factors(RuntimeContext &context) {
...
+        // Ignore if inactive or newborn or aged 19 and over.
         if (!person.is_active()) {
             continue;
         }
+        if (person.age == 0) {
+            continue;
+        }
+        if (person.age >= 19) {
+            continue;
+        }
```
dalonsoa, line 249:
```suggestion
        if (!person.is_active() || ((person.age == 0) || (person.age >= 19))) {
            continue;
        }
```

dalonsoa (approved):
Much better. The code is still complex, but the generate risk factors and update functions are more concise, clear and easier to follow :)

src/HealthGPS/kevin_hall_model.cpp
jamesturner246: Ah, that was @alexdewar's suggestion. What's the tie breaker!?

jamesturner246: I tend to agree, though. All those `if`s are a bit verbose.

alexdewar: I only suggested it because I thought it needed to be `if (person.age.has_value() && person.age != 0)`. If the whole thing fits on one line now, then that's def better!

src/HealthGPS/kevin_hall_model.cpp
```diff
@@ -73,118 +73,232 @@ void KevinHallModel::generate_risk_factors(RuntimeContext &context) {
...
     }
+}
 
-    // Intervention scenario: recieve adjustments from baseline scenario.
-    else {
-        auto message = context.scenario().channel().try_receive(context.sync_timeout_millis());
-        if (!message.has_value()) {
-            throw core::HgpsException(
-                "Simulation out of sync, receive Kevin Hall adjustments message has timed out");
+void KevinHallModel::update_non_newborns(RuntimeContext &context) const {
```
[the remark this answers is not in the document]
alexdewar: Haha, `clang-tidy` has also clocked that this is complex :stuck_out_tongue:

src/HealthGPS/kevin_hall_model.cpp
```diff
@@ -73,118 +73,232 @@ void KevinHallModel::generate_risk_factors(RuntimeContext &context) {
...
+            continue;
+        }
 
-        compute_bmi(person);
+        double W_power_mean = W_power_means.at(person.gender, person.age);
+        update_height(person, W_power_mean);
+    }
+}
+
+KevinHallAdjustmentTable KevinHallModel::recieve_weight_adjustments(RuntimeContext &context) const {
```
alexdewar, line 229:
```suggestion
KevinHallAdjustmentTable KevinHallModel::receive_weight_adjustments(RuntimeContext &context) const {
```

alexdewar: This is better than before, but I still think this would be better if it returned a `std::optional<KevinHallAdjustmentTable>`, with `std::nullopt` indicating that something went wrong. Then the caller can treat this as an error

alexdewar: You know what, scratch that. I think this is a perfectly reasonably way of doing it

jamesturner246:
Humph. Why all these builders doing strange stuff all of a sudden.

src/HealthGPS/kevin_hall_model.cpp
```diff
@@ -73,118 +73,208 @@ void KevinHallModel::generate_risk_factors(RuntimeContext &context) {
...
     }
+}
 
-    // Intervention scenario: recieve adjustments from baseline scenario.
-    else {
-        auto message = context.scenario().channel().try_receive(context.sync_timeout_millis());
-        if (!message.has_value()) {
-            throw core::HgpsException(
-                "Simulation out of sync, receive Kevin Hall adjustments message has timed out");
+void KevinHallModel::update_non_newborns(RuntimeContext &context) const {
```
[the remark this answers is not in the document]
jamesturner246: No, you're complex...

alexdewar (approved):
LGTM!

PR #274, opened 2023-11-19
Intervention scenario

### LINKED ISSUE #271: Add intervention update
This change is already complete, pending merging.

It adds the remainder of the intervention scenario code for the new Kevin Hall model.

jamesturner246:
This PR adds initialise and update code to `StaticLinearModel` for the intervention scenario. A policy is applied to each risk factor (here nutrients), which is based on the factor as if computed without intervention, plus a bounded linear model of a person's state which produces a change in that individual in the intervention scenario as a result of the intervention policy.

The code can now be considered feature-complete and ready for simulations, but very much untested to production standard.

Fixes #271.


jamesturner246:
Oooofffff........ I am done.

jamesturner246:
Okay... now done...

src/HealthGPS.Console/model_parser.cpp
```diff
@@ -141,59 +141,107 @@ std::unique_ptr<hgps::StaticLinearModelDefinition>
...
+            fmt::format("Risk factor count ({}) does not match intervention "
+                        "policy covariance matrix column count ({})",
+                        opt["RiskFactorModels"].size(), policy_covariance_table.num_columns())};
     }
 
-    // Compute Cholesky decomposition of correlation matrix.
-    auto cholesky = Eigen::MatrixXd{Eigen::LLT<Eigen::MatrixXd>{correlations}.matrixL()};
+    // Compute Cholesky decomposition of the intervention policy covariance matrix.
+    auto policy_cholesky =
+        Eigen::MatrixXd{Eigen::LLT<Eigen::MatrixXd>{policy_covariance}.matrixL()};
```
alexdewar, line 244: Are you using `policy_cholesky` or `cholesky` in this function? If not, you could skip casting the type here and instead of ``std::move`'ing it out of the function, just let it be copied, which will achieve the same thing.

example_new/Config.json
```diff
@@ -93,142 +93,15 @@
...
         ],
         "interventions": {
-            "active_type_id": null,
+            "active_type_id": "simple",
             "types": {
                 "simple": {
                     "active_period": {
                         "start_time": 2022,
-                        "finish_time": 2022
+                        "finish_time": null
```
alexdewar, line 101: Would it be clearer to represent "no finish time" with a missing key, rather than having it be `null`?

alexdewar (approved):
LGTM!

example_new/StaticLinear.json
dalonsoa, line 1: What's the difference between `Config.json` and `StaticLinear.json`? Why to files and what is each one used for?

src/HealthGPS.Console/model_parser.cpp
```diff
@@ -141,59 +141,107 @@ std::unique_ptr<hgps::StaticLinearModelDefinition>
...
+    // Check intervention policy covariance matrix column count matches risk factor count.
+    if (opt["RiskFactorModels"].size() != policy_covariance_table.num_columns()) {
         throw hgps::core::HgpsException{
-            fmt::format("Risk factor count ({}) does not match correlation "
-                        "matrix column count ({})",
-                        opt["RiskFactorModels"].size(), correlations_table.num_columns())};
+            fmt::format("Risk factor count ({}) does not match intervention "
+                        "policy covariance matrix column count ({})",
+                        opt["RiskFactorModels"].size(), policy_covariance_table.num_columns())};
     }
```
dalonsoa, line 240: So, out of curiosity, what should happen if a particular risk factor does not have any intervention in the current policy? In other words, if it should be left "untouched".

src/HealthGPS/static_linear_model.cpp
```diff
@@ -83,58 +101,133 @@ void StaticLinearModel::update_risk_factors(RuntimeContext &context) {
...
+        // Initialise residual.
+        auto residual_name = core::Identifier{names_[i].to_string() + "_policy_residual"};
+        double residual = residuals[i];
+        person.risk_factors[residual_name] = residual;
+
+        // Compute intervention policy.
+        double policy = linear[i] + residual;
+        policy = policy_ranges_[i].clamp(policy);
+        double factor_old = person.risk_factors.at(names_[i]);
+        double factor = factor_old * (1.0 + policy / 100.0);
```
dalonsoa, line 194: This is always computed, but only used if we are in the intervention scenario. Being that the case, should not this be within the conditional block below, so it is executed only in the intervention scenario?

src/HealthGPS/static_linear_model.cpp
```diff
@@ -83,58 +101,133 @@ void StaticLinearModel::update_risk_factors(RuntimeContext &context) {
...
+
+        // Apply intervention policy for intervention scenario.
+        if (scenario == ScenarioType::intervention) {
+            person.risk_factors.at(names_[i]) = factor;
+        }
     }
```
dalonsoa, line 225: As far as I can tell, this whole block (actually, the whole of `update_policies` function) only makes sense in the intervention scenario, so I would check if that is the case outside of the function, in the calling code, or just at the beginning of the function, returning immediately if we are not in an intervention scenario without doing anything.

dalonsoa (changes requested):
I've a couple of suggestions, but it looks ok, otherwise.

src/HealthGPS.Console/model_parser.cpp
jamesturner246: Presumably bad things... The workaround is to give a policy intercept of zero, with an empty coefficient array, but I agree that's not ideal. Once #231 is done, ideally intervention should become its own module, or just an instance of the static linear model. Maybe for now I can just autocomplete a no-op policy model, if no policy model is given. TLDR, bloated, needs refactoring.

src/HealthGPS/static_linear_model.cpp
jamesturner246: We have to be careful how much we `if (intervention)` out to keep the RNG of the baseline and intervention synchronised, but I agree that much of this can be chopped out in baseline.

src/HealthGPS/static_linear_model.cpp
jamesturner246: As above. I'll see what I can cut out.

src/HealthGPS.Console/model_parser.cpp
jamesturner246: i'm getting `std::bad_alloc` if I try to do this. I will leave it be for now.

alexdewar: Ah ok. Nvm then

jamesturner246:
I think most of these are now addressed, so I think it's time to merge this in finally, ready for new code.

src/HealthGPS/static_linear_model.cpp
```diff
@@ -58,6 +71,15 @@ void StaticLinearModel::generate_risk_factors(RuntimeContext &context) {
 
     // Adjust risk factors to match expected values.
     adjust_risk_factors(context, names_);
+
+    // Initialise everyone.
+    for (auto &person : context.population()) {
+        // HACK: start intervening after 2 years from sim start.
```
dalonsoa, line 77: Out of curiosity, why 2 and not 1 or 3 or something else?

src/HealthGPS/static_linear_model.cpp
dalonsoa: What's RNG?

src/HealthGPS/static_linear_model.cpp
```diff
@@ -83,81 +105,173 @@ void StaticLinearModel::update_risk_factors(RuntimeContext &context) {
...
+void StaticLinearModel::initialise_policies(Person &person, Random &random, bool intervene) const {
+    // NOTE: we need to keey baseline and intervention scenarios RNG in sync.
+
+    // Correlated residual sampling.
+    auto residuals = compute_residuals(random, policy_cholesky_);
+
+    // No-op if intervention scenario.
+    if (!intervene) {
+        return;
+    }
```
dalonsoa, line 190: Residuals are going to be used only it intervene is true, so why not checking this before? Actually, why running this function at all if intervene is false? I would make the check in the calling function.

I think I'm not totally understanding (or understanding at all) the synchronisation between scenarios and RNG.

dalonsoa (approved):
I have a couple of questions and, possibly, it will be useful to further break down some functions to ease their testing, but it is good enough as it is and I think it is more important to start tomorrow with a fresh repo.

src/HealthGPS/static_linear_model.cpp
jamesturner246: I think this was somewhat arbitrarily chosen by Ali, possibly to 'settle down' the simulation before intervening. This should eventually be customised. I'll open an issue.

src/HealthGPS/static_linear_model.cpp
jamesturner246: Sorry, I mean random number generators.

src/HealthGPS/static_linear_model.cpp
dalonsoa: Ahhh, ok that's very clear. Wow! That sounds incredibly brittle and inefficient if we need to run part of the same stuff in both scenarios even if things are not needed.

src/HealthGPS/static_linear_model.cpp
jamesturner246: #287

src/HealthGPS/static_linear_model.cpp
jamesturner246: No kidding. And the worst thing is that I'm not convinced it actually guarantees consistency. I can imagine a scenario where an intervention reduces mortality rate, resulting in different age distributions, so different state advancement no matter how synchronised you keep the RNGs. This needs some thought, but for now we're stuck with it!

Issue #284, opened 2024-01-16
Finalise India data and merge into mainline


jamesturner246:
There are a lot of iterations of the India dataset floating around, in varying degrees of completeness/correctness. 

It is time to settle on a 'final' version of the India dataset, and merge this into mainline. Note that we can still add fixes to this in future, but it is less confusing to have a 'single source of truth' which we should all use, unless otherwise noted.

This should hopefully mean that sharing experiments as we debug is a simple matter of sending a relevant JSON file.

The data will be contributed to healthgps in the form of a Github 'pull request'. I'm happy to walk you through the process,

jamesturner246:
Hi @jzhu20. Do you fancy giving this a go and seeing if we can get the India data in today? 

We should dig out the most up-to-date version of the India dataset we have -- I'm guessing there's a copy on the shared drive. Do you have a link to the most recent version? thanks!

jzhu20:
This is the most up-to-date India files with me. https://imperiallondon-my.sharepoint.com/:u:/r/personal/jjo11_ic_ac_uk/Documents/CHEPI/research/franco/Health-GPS_SHARED/Health-GPS_INDIA/india-files.7z?csf=1&web=1&e=I8SXVM
@jamesturner246 When are you available to guide me through it please? Thank you!

jamesturner246:
How's 10:00 tomorrow? Should have enough time to go through the whole GitHub PR process more generally then.

PR #292, opened 2024-01-18
Add India data files

jzhu20:
Add India data files from STOP;
Add India data files of ten cancers;
Add India data files of four new diseases (ischemic heart disease, ischemic stroke, intracerabral hemorrhage, subarachnoid hemorrhage);
Add relative risks of sodium on stomach cancer.


jamesturner246:
Just a note: it looks like `intercerebralhemorrhage` is also added in this change. The data seems to be present for India only.

data/analysis/disability_weights.csv
```diff
@@ -1,10 +1,35 @@
 disease,weight
 alzheimer,0.298333333
```
jamesturner246, line 2: The new diseases, including `ischemicheartdisease` are not present here. Should probably be added.

data/diseases/Metadata.json
```diff
@@ -0,0 +1,218 @@
+﻿{
+    "input_file": {
+        "running": {
+            "diseases": [
+                "alzheimer",
```
jamesturner246, line 5: Same here.

data/diseases/Metadata.json
```diff
@@ -0,0 +1,218 @@
...
+                "pancreascancer",
+                "stomachcancer",
+                "thyroidcancer"
+            ]
+        }
+    },
+    "index_file": {
+        "diseases": {
+            "registry": [
+                {
```
jamesturner246, line 45: Same here.

data/index.json
```diff
@@ -76,18 +76,126 @@
         },
         "registry": [
             { "group": "other", "id": "alzheimer", "name": "Alzheimer" },
```
jamesturner246, line 78: Same here. This is probably a more serious issue, since e.g. `ischemicheartdisease` cannot be considered in the simulation.

jamesturner246 (changes requested):
The gist of this review is that the new diseases do not seem to be listed in the config and metadata files. They should be added, otherwise the simulation cannot consider and compute them.

jamesturner246:
Fixes #284

jamesturner246 (approved):

PR #293, opened 2024-01-18
Fix: policy originals were not being saved.

jamesturner246:
It's just as you said, @dalonsoa. Very brittle...

After the below commit, policy residuals (initialised once and used throughout lifetime) were not being saved because I cut too much code out.

https://github.com/imperialCHEPI/healthgps/commit/10a07448f31960acfa5d3cba782067e76de3c008

This small PR fixes it again, and clarifies some of the comments.


dalonsoa (approved):
Oh my...

Issue #234, opened 2023-10-23
Do we really need to error check RF model params twice?


jamesturner246:
Right now they're checked twice: once in the model definition (factory) class, and again in the final risk factor model class.

This is unnecessary, besides I keep forgetting to copy the exact same test to both definition and model.

PR #304, opened 2024-02-01
Add dummy risk factor model class.

### LINKED ISSUE #300: Add a 'dummy' model type for constant risk factors
For this task, we will create a new 'dummy' risk factor model `RiskFactorDummyModel`, which will simply assign a given set of risk factors to some configurable constant value (perhaps mean by age and gender).

This will allow us to isolate the disease mechanism entirely (see #277) and, hopefully, will greatly speed up the diagnosis of unexpected disease behaviour for our 'minimal example' debug configurations, and eventually facilitate easier unit tests for risk factor models -- this dummy model can be considered a 'mock' risk factor model.

jamesturner246:
Fixes #300 

This change adds a simple `DummyModel` risk factor model, and `DummyModelDefinition` classes, along with `model_parser.cpp` code for loading parameters from a new `Dummy.json` config.

The idea is that we can force risk factors to certain configurable constant values, and modify them by some configurable constant amount in the intervention scenario.

This removes much of the stochasticity from the model, allowing us (and future unit tests) to focus on only effect and stochasticity in the disease mechanism, with minimal example code.

### DIFF
--- a/src/HealthGPS.Console/model_parser.cpp
+++ b/src/HealthGPS.Console/model_parser.cpp
@@ -81,6 +81,9 @@ std::unique_ptr<hgps::RiskFactorModelDefinition>
 load_static_risk_model_definition(const std::string &model_name, const poco::json &opt,
                                   const host::Configuration &config) {
     // Load this static model with the appropriate loader.
+    if (hgps::core::case_insensitive::equals(model_name, "dummy")) {
+        return load_dummy_model_definition(hgps::RiskFactorModelType::Static, opt);
+    }
     if (hgps::core::case_insensitive::equals(model_name, "hlm")) {
         return load_hlm_risk_model_definition(opt);
     }
@@ -331,6 +334,9 @@ std::unique_ptr<hgps::RiskFactorModelDefinition>
 load_dynamic_risk_model_definition(const std::string &model_name, const poco::json &opt,
                                    const host::Configuration &config) {
     // Load this dynamic model with the appropriate loader.
+    if (hgps::core::case_insensitive::equals(model_name, "dummy")) {
+        return load_dummy_model_definition(hgps::RiskFactorModelType::Dynamic, opt);
+    }
     if (hgps::core::case_insensitive::equals(model_name, "ebhlm")) {
         return load_ebhlm_risk_model_definition(opt, config);
     }
--- a/src/HealthGPS.Console/model_parser.cpp
+++ b/src/HealthGPS.Console/model_parser.cpp
@@ -59,6 +59,24 @@ hgps::RiskFactorSexAgeTable load_risk_factor_expected(const host::Configuration 
     return table;
 }
 
+std::unique_ptr<hgps::DummyModelDefinition>
+load_dummy_model_definition(hgps::RiskFactorModelType type, const poco::json &opt) {
+    MEASURE_FUNCTION();
+
+    // Get dummy model parameters
+    std::vector<hgps::core::Identifier> names;
+    std::vector<double> values;
+    std::vector<double> policy;
+    for (const auto &[key, json_params] : opt["ModelParameters"].items()) {
+        names.emplace_back(key);
+        values.emplace_back(json_params["Value"].get<double>());
+        policy.emplace_back(json_params["Policy"].get<double>());
+    }
+
+    return std::make_unique<hgps::DummyModelDefinition>(type, std::move(names), std::move(values),
+                                                        std::move(policy));
+}
+
 std::unique_ptr<hgps::RiskFactorModelDefinition>
 load_static_risk_model_definition(const std::string &model_name, const poco::json &opt,
                                   const host::Configuration &config) {
--- a/src/HealthGPS.Console/model_parser.cpp
+++ b/src/HealthGPS.Console/model_parser.cpp
@@ -81,13 +81,13 @@ std::unique_ptr<hgps::RiskFactorModelDefinition>
 load_static_risk_model_definition(const std::string &model_name, const poco::json &opt,
                                   const host::Configuration &config) {
     // Load this static model with the appropriate loader.
-    if (hgps::core::case_insensitive::equals(model_name, "dummy")) {
+    if (model_name == "dummy") {
         return load_dummy_model_definition(hgps::RiskFactorModelType::Static, opt);
     }
-    if (hgps::core::case_insensitive::equals(model_name, "hlm")) {
+    if (model_name == "hlm") {
         return load_hlm_risk_model_definition(opt);
     }
-    if (hgps::core::case_insensitive::equals(model_name, "staticlinear")) {
+    if (model_name == "staticlinear") {
         return load_staticlinear_risk_model_definition(opt, config);
     }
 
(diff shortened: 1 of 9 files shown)


src/HealthGPS.Console/model_parser.cpp
```diff
@@ -59,10 +59,31 @@ hgps::RiskFactorSexAgeTable load_risk_factor_expected(const host::Configuration
     return table;
 }
 
+std::unique_ptr<hgps::DummyModelDefinition>
+load_dummy_model_definition(hgps::RiskFactorModelType type, const poco::json &opt) {
```
dalonsoa, line 63: Let's go back to following best practices add a docstring here.

src/HealthGPS.Console/model_parser.cpp
```diff
@@ -59,10 +59,31 @@ hgps::RiskFactorSexAgeTable load_risk_factor_expected(const host::Configuration
...
+
+    return std::make_unique<hgps::DummyModelDefinition>(type, std::move(names), std::move(values),
+                                                        std::move(policy));
+}
+
 std::unique_ptr<hgps::RiskFactorModelDefinition>
 load_static_risk_model_definition(const std::string &model_name, const poco::json &opt,
                                   const host::Configuration &config) {
     // Load this static model with the appropriate loader.
+    if (hgps::core::case_insensitive::equals(model_name, "dummy")) {
```
dalonsoa, line 84: Would it be useful for readability purposes - and maybe even for performance, I don't know - to turn this into a switch statement?

```cpp
switch(model_name) {
    case "dummy":
        return load_dummy_model_definition(hgps::RiskFactorModelType::Static, opt);
    case "hlm":
        return load_hlm_risk_model_definition(opt);
    case "staticlinear":
        return load_staticlinear_risk_model_definition(opt, config);
    default:
         throw hgps::core::HgpsException{
             fmt::format("Static model name '{}' not recognised", model_name)};
         }
}
```

Also, I would require that the model name is properly provided as lowercase. Allowing to give "Staticlinear", "StaTicLinEar", "staticLinear", etc. is just a mess and affect readability. If we want to enable that, let's make a conversion lowercase once when reading the configuration from JSON and move on assuming only lowercase thereafter.

src/HealthGPS.Console/model_parser.cpp
```diff
@@ -59,10 +59,31 @@ hgps::RiskFactorSexAgeTable load_risk_factor_expected(const host::Configuration
...
+        values.emplace_back(json_params["Value"].get<double>());
+        policy.emplace_back(json_params["Policy"].get<double>());
+    }
+
+    return std::make_unique<hgps::DummyModelDefinition>(type, std::move(names), std::move(values),
+                                                        std::move(policy));
+}
+
 std::unique_ptr<hgps::RiskFactorModelDefinition>
 load_static_risk_model_definition(const std::string &model_name, const poco::json &opt,
```
dalonsoa, line 81: And, as we are updating this function, let's take the opportunity to add a docstring :smiley:

src/HealthGPS.Console/model_parser.cpp
```diff
@@ -313,6 +334,9 @@ std::unique_ptr<hgps::RiskFactorModelDefinition>
 load_dynamic_risk_model_definition(const std::string &model_name, const poco::json &opt,
```
dalonsoa, line 334: And the very same comments on the docstring, lowercase and changing to switch here.

src/HealthGPS/dummy_model.cpp
```diff
@@ -0,0 +1,75 @@
+#include "dummy_model.h"
```
dalonsoa, line 1: What convention are we following on adding docstrings in header files or code? I've seen [this comment](https://stackoverflow.com/a/3572839/3778792) making a distinction. I don't mind as long as things are commented.

src/HealthGPS.Console/model_parser.cpp
```diff
@@ -59,10 +59,31 @@ hgps::RiskFactorSexAgeTable load_risk_factor_expected(const host::Configuration
...
+
+    // Get dummy model parameters
+    std::vector<hgps::core::Identifier> names;
+    std::vector<double> values;
+    std::vector<double> policy;
+    for (const auto &[key, json_params] : opt["ModelParameters"].items()) {
+        names.emplace_back(key);
+        values.emplace_back(json_params["Value"].get<double>());
+        policy.emplace_back(json_params["Policy"].get<double>());
+    }
```
dalonsoa, line 74: This will fail if the three lists have different length. Is it worth to add some check or add some meaningful message if that happen, or we just leave it fail?

dalonsoa (changes requested):
Just a couple of comments, mostly related to readability. Otherwise, I think this is going to be really useful for debugging and testing purposes!

src/HealthGPS/dummy_model.cpp
jamesturner246: The convention I have been following here is to include docstrings in headers only, and to have inline comments in the CPP file code. Generally a matter of taste, but having them in both means needing to ensure that both are up to date, as opposed to just needing to keep header docstrings up to date. I'm open to being convinced otherwise, but that has struck me as a good compromise so far.

I see a few fn declarations in headers that have been missed. I'll make sure they're added.

src/HealthGPS.Console/model_parser.cpp
jamesturner246: The old docstrings in headers vs cpp files debate! See below.

src/HealthGPS.Console/model_parser.cpp
jamesturner246: > Would it be useful...

I agree it would tidy it up a bit. Unfortunately C++ has other ideas: https://stackoverflow.com/questions/650162/why-cant-the-switch-statement-be-applied-to-strings

Ultimately agree about case. I've bought it up before somewhere also. Not sure how big a job it is to go through all the code and remove case insensitivity. I'll make an issue.

src/HealthGPS.Console/model_parser.cpp
jamesturner246: Right, they are expected to be the same length. I'll add checks.

src/HealthGPS.Console/model_parser.cpp
dalonsoa: No preference at all - I don't know the common practices of C++. Having said that, the actual functionality is in the source code, not in the header, so I think I will find useful to have a description of the function alongside the code to fully understand what's going on.

jamesturner246: Like in the link you gave, have a 'developer' docstring in the .cpp and a 'user' docstring in the header? It's not a bad idea tbh. Bit of extra work, but I can see the use. Worth debating next meet.

src/HealthGPS.Console/model_parser.cpp
```diff
@@ -59,14 +59,35 @@ hgps::RiskFactorSexAgeTable load_risk_factor_expected(const host::Configuration
...
+    return std::make_unique<hgps::DummyModelDefinition>(type, std::move(names), std::move(values),
+                                                        std::move(policy));
+}
+
 std::unique_ptr<hgps::RiskFactorModelDefinition>
 load_static_risk_model_definition(const std::string &model_name, const poco::json &opt,
                                   const host::Configuration &config) {
     // Load this static model with the appropriate loader.
-    if (hgps::core::case_insensitive::equals(model_name, "hlm")) {
+    if (model_name == "dummy") {
```
dalonsoa, line 84: Why on Earth is C++ happy to make this comparison but unable to use a `switch` statement with strings? It does not make any sense!!

src/HealthGPS/dummy_model.cpp
```diff
@@ -0,0 +1,81 @@
...
+    }
+    if (values_.empty()) {
+        throw core::HgpsException("Risk factor values list is empty");
+    }
+    if (policy_.empty()) {
+        throw core::HgpsException("Risk factor policy lisy empty");
+    }
+    if (names_.size() != values_.size() || names_.size() != policy_.size()) {
+        throw core::HgpsException("Risk factor name, value and policy list sizes mismatch");
+    }
```
dalonsoa, line 74: This question applies equally to the other models, but why do we need to check this in both classes, the model class and the model definition class? Surely, it is impossible to have one without the other, right?

dalonsoa (approved):
Pending on the docstring conversation, this looks good!

src/HealthGPS/dummy_model.cpp
jamesturner246: You're right, it seems iffy. There's even an issue open about it: [#234](https://github.com/imperialCHEPI/healthgps/issues/234). Might as well just remove them now and be done with it.

src/HealthGPS.Console/model_parser.cpp
jamesturner246: https://xkcd.com/353/

PR #306, opened 2024-02-07
Remove testing with clang

dalonsoa:
As it is very annoying having that workflow systematically failing. Also, having a workflow that fails and is ignored defeats the purpose of having that workflow, to start with.

@jamesturner246 feel free to disagree and ignore this PR.


jamesturner246 (approved):
Agree. It is annoying. We were hoping in vain that the once-fine ubuntu clang builder would fix itself (see #237). 

I'll leave the issue open as a reminder to eventually re-add with a working clang builder, but this PR should be merged.

PR #307, opened 2024-02-08
Dummy policy start

jamesturner246:
This small change adds a per-factor configurable start iteration on which to apply intervention policy updating.

model_parser.cpp is added to load these values from a `DummyModel.json` config. Main changes are in `dummy_model.cpp`. Changes in `static_linear_model.cpp` are minor.

@TinyMarsh, I'll add you to the review if you want to get more of an idea on the workings, but no sweat if you don't.

### DIFF
--- a/src/HealthGPS.Console/model_parser.cpp
+++ b/src/HealthGPS.Console/model_parser.cpp
@@ -67,14 +67,16 @@ load_dummy_model_definition(hgps::RiskFactorModelType type, const poco::json &op
     std::vector<hgps::core::Identifier> names;
     std::vector<double> values;
     std::vector<double> policy;
+    std::vector<int> policy_start;
     for (const auto &[key, json_params] : opt["ModelParameters"].items()) {
         names.emplace_back(key);
         values.emplace_back(json_params["Value"].get<double>());
         policy.emplace_back(json_params["Policy"].get<double>());
+        policy_start.emplace_back(json_params["PolicyStart"].get<int>());
     }
 
     return std::make_unique<hgps::DummyModelDefinition>(type, std::move(names), std::move(values),
-                                                        std::move(policy));
+                                                        std::move(policy), std::move(policy_start));
 }
 
 std::unique_ptr<hgps::RiskFactorModelDefinition>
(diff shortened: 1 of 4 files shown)


src/HealthGPS.Console/model_parser.cpp
```diff
@@ -67,14 +67,16 @@ load_dummy_model_definition(hgps::RiskFactorModelType type, const poco::json &op
     std::vector<hgps::core::Identifier> names;
     std::vector<double> values;
     std::vector<double> policy;
+    std::vector<int> policy_start;
     for (const auto &[key, json_params] : opt["ModelParameters"].items()) {
         names.emplace_back(key);
         values.emplace_back(json_params["Value"].get<double>());
         policy.emplace_back(json_params["Policy"].get<double>());
+        policy_start.emplace_back(json_params["PolicyStart"].get<int>());
```
dalonsoa, line 75: Just to confirm I get this right, you need a start value, in years since the start of the simulation, for each of the risk factors, so the policy is applied from that point. Is this correct?

src/HealthGPS/dummy_model.cpp
```diff
@@ -27,26 +28,32 @@ void DummyModel::update_risk_factors(RuntimeContext &context) {
...
         }
 
-        set_risk_factors(entity, context.scenario().type());
+        set_risk_factors(entity, context);
     }
 }
 
-void DummyModel::set_risk_factors(Person &person, ScenarioType scenario) const {
-    for (auto i = 0u; i < names_.size(); ++i) {
+void DummyModel::set_risk_factors(Person &person, RuntimeContext &context) const {
```
dalonsoa, line 35: I would pass `person`, `scenario` and `time_elapsed`. That;s easier to work with (to understand and test) than passing the whole context. It is also more efficent as you only need to calculate such a time once instead of once per risk factor.

dalonsoa (approved):
I've just included a suggestion, but otherwise it looks good!

src/HealthGPS.Console/model_parser.cpp
jamesturner246: Correct. I'll clarify in the docstring.

src/HealthGPS/dummy_model.cpp
jamesturner246: Done!

Issue #289, opened 2024-01-16
Unit tests for linear model methods in StaticLinear


jamesturner246:
We need unit tests for the linear models which compute the base risk factors in `static_linear_model.cpp` - i.e. `initialise_factors` and `update_factors`.

These should check:
- [ ] are residuals computed and stored correctly? (init and update)
- [ ] does residual correlation match provided correlation matrix?
- [ ] is the 'information speed' mechanism from old residual to new computed correctly?
- [ ] is the linear component computed and saved correctly?
- [ ] is the inverse Box Cox computed correctly?
- [ ] is RNG state of baseline and intervention still equal afterwards?

TinyMarsh:
Hey @jamesturner246 (or anyone else who can help!), I have a question which is probably more to do with general best practice with testing in C++ than it is specifically to do with creating these specific tests.

So initially I am looking at testing the `StaticLinearModel::initialise_factors` method. This has two parameters; a `Person` type, and a `Random` type. These are easily instantiated for test purposes, but I'm wondering the best way to go about instantiating a test `StaticLinearModel` instance so that I can call the `initialise_factors` method on it.

The constructor for `StaticLinearModel` takes several parameters. Is the best way to do this to just go through each of these parameters and create an instance of each to use as arguments for the constructor so that I can create an instance, or is there a better way to do this? It seems like we should have a way to create tests instances of `StaticLinearModel`.

Sorry if this is a naive question, I just wanted to make sure I didn't waste time doing this the "wrong" way in case there's a better way.

jamesturner246:
No worries, you basically hit the nail on the head. As a 'fixture', we can construct some instance of static linear with our own parameters which make checking simpler.

So maybe have all/some of the tests use some generic fixture `StaticLinear` which has some fake vector of coefficients, cov matrix, etcetera, passed in via the ctor, that's only complicated enough to do the tests we want, and no more complicated. I realise that's not an amazingly concise answer, so happy to chat over Teams.

TinyMarsh:
Thanks @jamesturner246. This is what I needed to hear and points me in the right direction so thank you!

TinyMarsh:
Hi @jamesturner246, I'm getting a bit stuck with this so it might be good if we had a chat actually. I think I just need my head pointing in the right direction.

[...]

PR #316, opened 2024-02-19
Initial pass at implementing test fixtures

TinyMarsh:
Some of the models/policy models have been implemented. I am having difficulty with how to instantiate the correlation matrices so would appreciate some help there.

Many of the parameters are remaining as a todo. Addresses #289

### DIFF
--- a/src/HealthGPS.Tests/StaticLinearModel.Test.cpp
+++ b/src/HealthGPS.Tests/StaticLinearModel.Test.cpp
@@ -1,6 +1,7 @@
 #include "pch.h"
 
 #include "HealthGPS/static_linear_model.h"
+#include "HealthGPS/mtrandom.h"
 
 // Create test fixture for a StaticLinearModel class instance
 class StaticLinearModelTestFixture : public::testing::Test {
@@ -17,9 +18,9 @@ class StaticLinearModelTestFixture : public::testing::Test {
         std::vector<hgps::core::DoubleInterval> policy_ranges;
         Eigen::MatrixXd policy_cholesky;
 
-        hgps::StaticLinearModel* testModel;
-        hgps::Person* testPerson;
-        hgps::Random* testRandom;
+        hgps::StaticLinearModel* test_model;
+        hgps::Person* test_person;
+        hgps::Random* test_random;
 
         void SetUp() override {
             auto expected = hgps::RiskFactorSexAgeTable{};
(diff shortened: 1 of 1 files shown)


src/HealthGPS.Tests/StaticLinearModel.Test.cpp
```diff
@@ -0,0 +1,190 @@
...
+        std::vector<hgps::LinearModelParams> policy_models;
+        std::vector<hgps::core::DoubleInterval> policy_ranges;
+        Eigen::MatrixXd policy_cholesky;
+
+        void SetUp() override {
+            auto expected = hgps::RiskFactorSexAgeTable{};
+            std::vector<double> factorsMale = {200.,40.,30.,2000.,1.,1000.,3.,50.};
+            std::vector<double> factorsFemale = {200.,40.,30.,2000.,1.,1000.,3.,50.};
+
+            std::unordered_map<hgps::core::Identifier, std::vector<double>> factorsMale = {
```
jamesturner246, line 24: the vector of means by age needs a different name than the container. Maybe `factor_means_male` or something (also snake-case for consistency).

src/HealthGPS.Tests/StaticLinearModel.Test.cpp
```diff
@@ -0,0 +1,190 @@
...
+            expected.emplace_row(hgps::core::Gender::female, factorsFemale);
+
+            std::vector<hgps::core::Identifier> names = {"FoodCarbohydrate", "FoodFat"};
+
+            // Define models
+            // We will define 2 RiskFactorModels: carb_model and fat_model 
+            std::vector<hgps::LinearModelParams> models;
+            hgps::LinearModelParams carb_model;
+            carb_model.intercept = -0.241505734056409;
+            std::unordered_map<hgps::core::Identifier, double> coefficients = {
```
jamesturner246, line 44: These coefficient vectors can be initialised in-place, and you won't need to use the extra `coefficients` vector.
```cpp
           carb_model.coefficients = {
```

src/HealthGPS.Tests/StaticLinearModel.Test.cpp
```diff
@@ -0,0 +1,190 @@
...
+        Eigen::MatrixXd policy_cholesky;
+
+        void SetUp() override {
+            auto expected = hgps::RiskFactorSexAgeTable{};
+            std::vector<double> factorsMale = {200.,40.,30.,2000.,1.,1000.,3.,50.};
+            std::vector<double> factorsFemale = {200.,40.,30.,2000.,1.,1000.,3.,50.};
+
+            std::unordered_map<hgps::core::Identifier, std::vector<double>> factorsMale = {
+                {hgps::core::Identifier{"0"}, factorsMale},
+                {hgps::core::Identifier{"1"}, factorsMale}
```
jamesturner246, line 26: These identifiers will need to be the same as the factors in `names`, so `FoodCarbohydrate`, etc.

jamesturner246 (changes requested):
Thanks @TinyMarsh. Looks like it's taking shape!

Some minor notes about initialisation, name convention and model name, but on the right track :+1: 

Happy to take another look later on.

jamesturner246:
Sorry, just saw the fixture class. I think what you'll find is that you need to initialise all the vectors and tables and stuff above *inside* the fixture constructor, above where you construct the staticlinearmodel, as I don't *think* there's a way to pass in arguments to fixtures.

You'll also need a variable inside the fixture class to actually store the `StaticLinearModel`.

jamesturner246:
Looking good so far. I think the `FRIEND_TEST` way is probably the way to go whilst the model is bloated like this. Eventually we can separate it out enough so as not to need i, but good for now. 

A couple of points:

1. in these more modern C++ versions, we basically avoid `new`/`delete` entirely in favour of smart pointers, with e.g. `std::make_unique` which returns `std::unique_ptr` and `std:make_shared` which returns a `std:unique_ptr`. Avoids memory leaks among other things. If it can't be allocated in the stack frame, use smart pointers. Never `new`/`delete`, unless you're fiddling and making your own smart pointes.
2. Are the `pre-commit` hooks working for you? Could you try `pre-commit install` again? 

Thanks!

PR #333, opened 2024-03-08
Update RR based on hack_income_class

jzhu20:
Update relative risks of BMI on ischemic heart disease, ischemic stroke, intracerebral hemorrhage, subarachnoid hemorrhage, in hack_income_class branch.


jamesturner246 (approved):
LGTM. I don't see anything obviously wrong. Go ahead.

p.s. I suppose you've ran a model on these values to check they work and produce reasonable outputs?

PR #346, opened 2024-04-30
Hack income class hslope threads

jamesturner246:
Merging TBB threads into hack income class branch.


jamesturner246:
The Windows `getenv` weirdness is unfortunately an issue, but we will deal with that when we merge to `main`.

PR #348, opened 2024-05-02
TBB'ification of threading and tidy-up

alexdewar:
As discussed, we want to try to bring as many of the threads in Health-GPS under the control of TBB's task scheduler so that the maximum thread limit we impose on it will be respected program-wide.

To do this, I've changed `EventMonitor` to use a `tbb::task` group for the info and results threads, which previously ran on `std::jthread`s (which, oddly enough, were housed in a `std::vector`, despite there only ever being two of them). While I was at it, I noticed that the code was doing a lot of waiting around for messages to be received (using `std::this_thread::sleep_for`), which is likely to be inefficient, so I switched out the `tbb::concurrent_queue`s we were using for `tbb::concurrent_bounded_queue`s, which provide blocking methods.

I also removed the pointless additional thread that was being created to run the baseline simulation, so it just runs on the main thread.

These changes mean that on my 4-core Intel laptop there are now only 9 threads created by default (== 8 logical cores + 1) rather than 12, which is what it was before. I still don't know where that ninth thread is coming from -- maybe TBB creates 8 threads of its own and then the ninth one is just the main thread? In any case, it's an improvement and hopefully it'll stop @jzhu20's jobs from being killed.

Pleasingly, these changes also seem to bring a modest performance increase (>10%).

Previously on the `example_new` dataset it took:

```
________________________________________________________
Executed in  367.99 secs    fish           external
   usr time  773.88 secs    1.41 millis  773.88 secs
   sys time   22.55 secs    0.00 millis   22.55 secs
```

Now it takes:

```
________________________________________________________
Executed in  320.82 secs    fish           external
   usr time  672.96 secs  420.00 micros  672.96 secs
   sys time   13.99 secs   87.00 micros   13.99 secs
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,6 +48,9 @@ if(MSVC)
     if(WARNINGS_AS_ERRORS)
         add_compile_options(/WX)
     endif()
+
+    # Suppress warnings about std::getenv
+    add_compile_definitions(-D_CRT_SECURE_NO_WARNINGS)
 else()
     add_compile_options(-Wall -Wextra -Wpedantic)
 
--- a/src/HealthGPS.Console/event_monitor.cpp
+++ b/src/HealthGPS.Console/event_monitor.cpp
@@ -9,7 +9,7 @@
 
 namespace host {
 EventMonitor::EventMonitor(hgps::EventAggregator &event_bus, ResultWriter &result_writer)
-    : result_writer_{result_writer} {
+    : result_writer_{result_writer}, tg_{tg_context_} {
     handlers_.emplace_back(event_bus.subscribe(hgps::EventType::runner, [this](auto &&PH1) {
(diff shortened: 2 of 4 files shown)


alexdewar:
Ok, so on reflection, I think the cleanup code for the threads is now not correct. I've unpushed the last commit (the `tbb::concurrent_bounded_queue` stuff), so this branch now just contains the minimal amount of code needed to get rid of all the `std::jthread`s without changing the overall structure, which I think is probably all we should merge for now. I'd hate for @jzhu20's jobs to end up hanging after 8 hours.

I think changing over to `tbb::concurrent_bounded_queue`s is the right thing to do though, so I'll open an issue for it and we can do it later.

jamesturner246:
Thanks @alexdewar. Mind having a very quick chat o go over this?

jamesturner246 (approved):
LGTM!

Issue #349, opened 2024-05-02
Bounds in nutrients


jzhu20:
Update bounds in nutrient intake, without filtering outliers.

jzhu20:
Min and max values of these four nutrients are

- Carbohydrate: 22.09977, 1979.988
- Fat: 1.093948, 561.4858
- Protein: 3.649882, 309.119
- Sodium: 9.986686, 45379.46

jamesturner246:
Wow, they're considerably wider. Thanks -- I'll merge them in.

Issue #281, opened 2024-01-11
Null value in nutrients under intervention


jzhu20:
Null values are found in **_foodcarbohydrate, foodfat, foodprotein, foodsodium, energy intake, weight, bmi_**, under intervention, in some specific years. I'm not sure if it's an issue. For example,
### Female
Null values in 2037, 2039, 2040
### Male
Null values in 2030, 2036, 2050, 2051, 2055
### Female
Null values in 2030, 2036, 2037, 2040, 2043
### Male
Null values in 2028, 2029, 2051

jamesturner246:
Is this from the same simulation as #277?

jamesturner246:
Also, could you give a bit of info about where these NaN are appearing? Are they in the output csv file, or the json file?

Could you upload the file these appear?

jzhu20:
All are based on the same simulation. NaN appeared in output csv files. It can be found in this sharefolder.
https://imperiallondon-my.sharepoint.com/:f:/r/personal/jjo11_ic_ac_uk/Documents/CHEPI/research/franco/Health-GPS_SHARED/Health-GPS_INDIA/Visualization?csf=1&web=1&e=RQOLaa
For example, in HealthGPS_Result_S1.csv, lines 3630, 6056, 6461, etc.

jamesturner246:
Okay. I suggest we make a minimal example which produces these NaN results. Small a population and runtime as possible, with no other cruft to distract us. Can talk about it tomorrow.

jzhu20:
No NaN in France example results.

jamesturner246:
Great, thanks. Need to figure out if it's still happening with India data now.

jzhu20:
No NaN in results with India/France inputs and India diseases data.

jamesturner246:
Ah fantastic, thanks. Can this issue be considered resolved now? Or is there some other permutation of data which is problematic?

jzhu20:
Shall we wait until we get results using the new KH model? There might be problems there.

jzhu20:
I got "nan" in simulation results using KH model. @jamesturner246 
Results can be found here.
[...]

PR #353, opened 2024-05-03
Main hslope nutrient bounds

jamesturner246:
Changed static linear model code to place bounds on nutrient values to be prevent NaNs due to negative nutrients.

Also changed the expected value adjustment code to return NaN mean for empty baseline groups instead of zero, and intervention groups are *not* adjusted if baseline group mean is NaN. This prevents a very large `delta = expected - sim_mean` if baseline group is empty, since `sim_mean` for empty groups is now NaN, not zero.

Addresses #281 and #349.


src/HealthGPS/static_linear_model.cpp
```diff
@@ -112,7 +112,12 @@ void StaticLinearModel::initialise_factors(Person &person, Random &random) const
         double expected = get_risk_factor_expected().at(person.gender, names_[i]).at(person.age);
         double factor = linear[i] + residual * stddev_[i];
         factor = expected * inverse_box_cox(factor, lambda_[i]);
-        person.risk_factors[names_[i]] = factor;
+
+        // Clamp risk factor to range.
+        double factor_clamped = ranges_[i].clamp(factor);
+
+        // Save risk factor.
+        person.risk_factors[names_[i]] = factor_clamped;
```
dalonsoa, line 120: Can this be simplified as..

```suggestion
        // Save risk factor.
        person.risk_factors[names_[i]] = ranges_[i].clamp(factor);
```
?

src/HealthGPS/static_linear_model.cpp
```diff
@@ -138,7 +143,12 @@ void StaticLinearModel::update_factors(Person &person, Random &random) const {
         double expected = get_risk_factor_expected().at(person.gender, names_[i]).at(person.age);
         double factor = linear[i] + residual * stddev_[i];
         factor = expected * inverse_box_cox(factor, lambda_[i]);
-        person.risk_factors.at(names_[i]) = factor;
+
+        // Clamp risk factor to range.
+        double factor_clamped = ranges_[i].clamp(factor);
+
+        // Save risk factor.
+        person.risk_factors.at(names_[i]) = factor_clamped;
```
dalonsoa, line 151: And here

src/HealthGPS/static_linear_model.cpp
```diff
@@ -171,8 +181,11 @@ void StaticLinearModel::initialise_policies(Person &person, Random &random, bool
         double factor_old = person.risk_factors.at(names_[i]);
         double factor = factor_old * (1.0 + policy / 100.0);
 
-        // Apply intervention policy.
-        person.risk_factors.at(names_[i]) = factor;
+        // Clamp policy-adjusted risk factor to range.
+        double factor_clamped = ranges_[i].clamp(factor);
+
+        // Save risk factor.
+        person.risk_factors.at(names_[i]) = factor_clamped;
```
dalonsoa, line 188: And here

src/HealthGPS/static_linear_model.cpp
```diff
@@ -200,8 +213,11 @@ void StaticLinearModel::update_policies(Person &person, bool intervene) const {
...
```
dalonsoa, line 220: And here

dalonsoa (changes requested):
Some tests are failing, but keeping that aside, I have a couple of comments and questions:

- This is clamping all factors, not just nutrients, right?
- Assuming negative nutrient values are related to malnutrition, why is that resulting in NaNs?

alexdewar (approved):
I've got concerns about how many input files are being changed in this PR, but as long as nothing's been broken, I think it's fine.

I've made one suggestion about naming but feel free to ignore. Otherwise looks good!

example_new/Config.json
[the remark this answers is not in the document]
jamesturner246: The only changes in StaticLinear.json are the `Range` fields in the `RiskFactorModels`. I merged in the other scenarios since they needed it straight away. Sorry, I should have added a note.

alexdewar:
PS -- my PR has a fix for the Windows issue. I've just merged it, so if you merge/rebase the tests should pass

src/HealthGPS/risk_factor_adjustable_model.cpp
```diff
@@ -77,9 +73,18 @@ void RiskFactorAdjustableModel::adjust_risk_factors(
...
-            person.risk_factors.at(factor) += delta;
+        for (size_t i = 0; i < factors.size(); i++) {
+            double delta = adjustments.at(person.gender, factors[i]).at(person.age);
+            double value = person.risk_factors.at(factors[i]) + delta;
+
+            // Clamp value to an optionally specified range.
+            if (ranges.has_value()) {
+                const auto &range = ranges.value().get()[i];
+                value = range.clamp(value);
+            }
```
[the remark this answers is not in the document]
jamesturner246: Only the risk factors that have models defined in StaticLinear.json, which all have `Range`s defined.

src/HealthGPS/static_linear_model.h
```diff
@@ -26,11 +26,12 @@ class StaticLinearModel final : public RiskFactorAdjustableModel {
     /// @param expected The risk factor expected values by sex and age
     /// @param names The risk factor names
     /// @param models The linear models used to compute a person's risk factor values
+    /// @param ranges The risk factor value boundaries
     /// @param lambda The lambda values of the risk factors
     /// @param stddev The standard deviations of the risk factors
     /// @param cholesky Cholesky decomposition of the risk factor correlation matrix
     /// @param policy_models The linear models used to compute a person's intervention policies
-    /// @param policy_ranges The boundaries of the intervention policy values
+    /// @param policy_ranges The intervention policy value boundaries
```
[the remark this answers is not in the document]
jamesturner246: I'll change it to range.

dalonsoa (approved):

PR #359, opened 2024-05-09
KevinHall `height_slope_` param for both females and males

jamesturner246:
While this PR is quite large, due to much of the recent stuff being merged into here, the only part to look at in this merge is changing `height_slope_` in `kevin_hall_model.cpp` from a single double to a map of doubles for female/male.

Search for text `slope` to find the changes, including the `HeightSlope` parameters in `KevinHall.json` and associated loading code in `model_parser.cpp`, as well as the changes to `kevin_hall_model.cpp` themselves.


dalonsoa (changes requested):
This is really not a good way of creating a PR. As far as I can tell, you are mixing things unrelated with each other and unrelated with the topic of the PR, modifying tons of files in the process.

You really need to be more selective and have a more rigorous approach when adding contributions. If this PR builds on work done in a different branch, then open the PR against that branch. If it does not, then break the work into smaller, conceptually related PRs form different branches.

We need to be an example to follow when contributing to a software project, and this not a good example of that.

PR #367, opened 2024-05-10
KevinHall `height_slope_` param for both females and males

jamesturner246:
this merge is changing `height_slope_` in `kevin_hall_model.cpp` from a single double to a map of doubles for female/male.

Search for text slope to find the changes, including the `HeightSlope` parameters in `KevinHall.json` and associated loading code in `model_parser.cpp`, as well as the changes to `kevin_hall_model.cpp` themselves.

PR #368, opened 2024-05-10
Replace std foreach for TBB foreach

jamesturner246:
Merge *AFTER* #367

`std:for_each` has no mechanism to enforce a thread limit -- a problem on HPC.

This change simply replaces each instance of `std::for_each` with the TBB equivalent `tbb::parallel_for`, allowing us to control the thread limit with TBB thread configuration.

Users can do this by setting the environment variable `OMP_THREAD_LIMIT`. Some threads remain, but will be reduced in another PR.


src/HealthGPS.Core/thread_util.h
```diff
@@ -22,34 +21,30 @@ template <class F, class... Ts> auto run_async(F &&action, Ts &&...params) {
...
 /// @tparam T Container type
 /// @tparam UnaryPredicate Predicate type
-/// @param policy The execution policy to use
 /// @param data The container to search
 /// @param pred The predicated object to apply
 /// @return A std::vector<size_t> with zero-based index of all occurrences.
-template <class Policy, class T, class UnaryPredicate>
-auto find_index_of_all(Policy &&policy, const T &data, UnaryPredicate pred) {
+template <class T, class UnaryPredicate>
+auto find_index_of_all(const T &data, UnaryPredicate pred) {
```
alexdewar, line 44: Are we sure that the policy is always to run in parallel here?

src/HealthGPS.Tests/Channel.Test.cpp
```diff
@@ -234,7 +235,8 @@ TEST(ChannelTest, MultipleConsumers) {
     });
 
     producer.join();
-    std::for_each(threads.begin(), threads.end(), [](std::jthread &thread) { thread.join(); });
+    tbb::parallel_for_each(threads.begin(), threads.end(),
+                           [](std::jthread &thread) { thread.join(); });
```
alexdewar, line 239: You don't want to run this in parallel. It could just be a ranged `for`-loop:

```suggestion
    for (auto &thread : threads) {
        thread.join();
    }
```

src/HealthGPS.Core/thread_util.h
alexdewar, line 1: Will the "policy" for these functions always be parallel? If not, you'll need to handle the case where the desired policy is serial.

src/HealthGPS/disease.cpp
```diff
@@ -68,24 +69,23 @@ std::unique_ptr<DiseaseModule> build_disease_module(Repository &repository,
...
+
+        auto &disease_definition = repository.get_disease_definition(info, config);
+        auto &lms_definition = repository.get_lms_definition();
+        auto classifier = WeightModel{LmsModel{lms_definition}};
+
+        // Sync region
+        std::scoped_lock lock(m);
+        models.emplace(core::Identifier{info.code},
+                       registry.at(info.group)(disease_definition, std::move(classifier),
+                                               config.settings().age_range()));
```
alexdewar, line 87: I have a feeling this locking is going to hammer performance here... Another option would be to use one of TBB's fancy lockless data structures instead of a regular `std::map`, but it's probably not worth the effort.

src/HealthGPS/kevin_hall_model.cpp
```diff
@@ -629,7 +631,7 @@ KevinHallAdjustmentTable KevinHallModel::compute_mean_weight(Population &populat
 
         double value;
         if (power.has_value()) {
-            value = pow(person.risk_factors.at("Weight"_id), power.value());
+            value = pow(person.risk_factors.at("Weight"_id), power.value()[person.gender]);
```
alexdewar, line 634: Is this change supposed to be in this branch?

alexdewar (approved):
A few small comments and a suggestion, but otherwise good.

I think you did inadvertently bring in some changes from your other branch (the gender-specific height stuff), but not many.

alexdewar:
Sorry, just saw your comment about reviewing after #367. If it were me, I'd have marked this as a draft PR and only tagged reviewers after the other one was merged... But maybe I should actually read people's PR descriptions :laughing:

CMakeLists.txt
```diff
@@ -48,6 +48,9 @@ if(MSVC)
     if(WARNINGS_AS_ERRORS)
         add_compile_options(/WX)
     endif()
+
+    # Suppress warnings about std::getenv
+    add_compile_definitions(-D_CRT_SECURE_NO_WARNINGS)
```
dalonsoa, line 53: Why is this necessary?

dalonsoa (approved):
The code looks ok to me, so all good from that side. 

I would, however, document somewhere why using the lower level TBB is necessary. Maybe a docstring somewhere or in the developer's documentation - if there's such a thing. Otherwise, future developers my wonder why.

Also, I've noticed we are using `oneapi`, which I guess we were using already, and the GH Actions are not complaining, so probably is not an issue, but just checking that this will no add any complication for the installation of Health-GPS by future users or in other systems.

src/HealthGPS.Core/thread_util.h
jamesturner246: Yeah, probably. oneTBB enforces serial when num_threads = 1, so you get both for free ([see here](person.gender)).

> With max_allowed_parallelism set to 1, global_control enforces serial execution of all tasks by the application thread(s), that is, the task scheduler does not allow worker threads to run. There is one exception: if some work is submitted for execution via task_arena::enqueue, a single worker thread will still run ignoring the max_allowed_parallelism restriction.

By the way, side note: I wonder if the exception above is what is causing the extra thread even after your TBB changes in the other PR?

src/HealthGPS/disease.cpp
jamesturner246: This part is only done in initialisation, and there's probably only about 30 or so diseases. Probably not worth bothering, since we might be re-writing this bit anyway,

CMakeLists.txt
jamesturner246: This was added because the Windows CI builder claims that `std::getenv` is unsafe. It isn't, in this case, since we are just reading and immediately using the thread limit env variable.

jamesturner246:
Added a note in `docs/development.md`.

The 'official' instructions for building and using hgps make use of cmake/vcpkg, which pulls in all the required packages automatically, and it seems unlikely this will change any time soon, so it should be okay for all but the weirdest use cases.

PR #371, opened 2024-05-14
New inputs

jamesturner246:
These are just new input data in `example_new` and data backend in `data`. No code changes.

No need to look at all the values (unless you really want to)...


dalonsoa (approved):
They look like nice values 😄 

LGTM!

alexdewar (approved):
LGTM!

Where did these data come from, out of interest? As a side note, maybe we should have the provenance of data either in the metadata somewhere or the documentation....

I haven't looked through all 200k lines of changes, but I assume they're fine :wink:

jamesturner246:
@dalonsoa Glad you agree they are fine looking values! :stuck_out_tongue_winking_eye: 

@alexdewar A bit of a mish-mash of different sources, some official and some not. Once we get the data sources code up, should make it easier to automatically include this metadata.

PR #375, opened 2024-05-16
Bounded nutrients

jamesturner246:
Last one to complete merge. :partying_face: 

Apply clamping to risk factors (nutrients), loaded from the `StaticLinear.json` config file to keep them in range.


dalonsoa (commented):
Looks good to me!

Issue #382, opened 2024-05-23
Move example projects to healthgps-examples repo


jamesturner246:
A number of job-dependent config and data files are scattered throughout the healthgps repo now, in the `example` and `example_new` directories.

For this change, we will move these examples into the new `healthgps-examples` repository, in some sensible layout, with some documentation.

PR #388, opened 2024-05-28
Move examples

jamesturner246:
The example project directories contain large JSON/CSV files which are subject to frequent change. 

This PR removes these example projects, and relocates them to the [healthgps examples repository](https://github.com/imperialCHEPI/healthgps-examples).

* `example` directory is now hosted at https://github.com/imperialCHEPI/healthgps-examples/tree/main/example_HLM_France
* `example_new` is now hosted at https://github.com/imperialCHEPI/healthgps-examples/tree/main/example_KevinHall_India

Documentation has been updated with the new links.

Addresses #382.


alexdewar (approved):
Happy to see these ma-hoosive files go.

I'm not entirely sure about the layout of the examples repo though. I feel you could drop the `example_*` prefix for all the folder names to make things a bit tidier. Then you could either put everything in an `examples` subfolder or just leave them in the root folder (the repo only contains examples so I think it's sort of implied that the subfolders will contain examples).

PR #390, opened 2024-05-29
Enable dependabot updates for GitHub Actions

alexdewar:
Let's keep the versions of Github Actions that we're using with dependabot.

Sadly dependabot doesn't (yet) support vcpkg, but maybe it will in future!


jamesturner246 (approved):
LGTM :+1:

Issue #366, opened 2024-05-10
Migrate input data to separate repository


alexdewar:
It should be decoupled from the Health-GPS version.

Issue #364, opened 2024-05-10
Allow for reading data from URL


alexdewar:
Depends on #363.

Rather than using a local directory for data files (in practice, the ones bundled with Health-GPS) it would be better if users could specify a URL to download the data from. This would allow us to decouple the version of these data files from the version of Health-GPS and would allow for better reproducibility (currently, I don't think there's a way to tell from the output data what the input data was).

PR #391, opened 2024-05-29
Read zipped input data

### LINKED ISSUE #363: Allow for reading data (and configs?) from zip file
This is a prerequisite to being able to download datasets via a URL (#364).

There are two ways this could be implemented:

1. We could read the files directly from the zip file
2. We could extract the zip file then use the existing code path for reading the files from a directory

alexdewar:
This PR adds the option to provide the input data as a path to a zip file rather than to a directory, e.g.:

```
HealthGPS.Console -f example_new/Config.json -s data_file.zip
```

Note that the `index.json` file must be in the root folder of the zip file otherwise it won't work.

While this feature isn't particularly useful by itself, it's a necessary step to being able to automatically download input data from a URL (#364).

I haven't added tests yet because the input files are likely to be moved out of this repo soon (#366), so I'll wait until we know what the final project structure looks like first. I didn't want to commit a zip file to the repo either if its contents are likely to change soon.

I've also broken out the schema-related code into its own `.cpp` file as `datamanager.cpp` was getting a bit bloated.

Closes #363.

### DIFF
--- a/src/HealthGPS.Datastore/CMakeLists.txt
+++ b/src/HealthGPS.Datastore/CMakeLists.txt
@@ -1,12 +1,21 @@
 find_package(fmt CONFIG REQUIRED)
 find_package(jsoncons CONFIG REQUIRED)
+find_package(libzippp CONFIG REQUIRED)
 
 add_library(HealthGPS.Datastore STATIC "")
 target_compile_features(HealthGPS.Datastore PUBLIC cxx_std_${CMAKE_CXX_STANDARD})
 
-target_sources(HealthGPS.Datastore PRIVATE "api.h" "datamanager.cpp" "datamanager.h" "schema.cpp"
-                                           "schema.h")
+target_sources(
+    HealthGPS.Datastore
+    PRIVATE "api.h"
+            "datamanager.cpp"
+            "datamanager.h"
+            "schema.cpp"
+            "schema.h"
+            "zip_file.cpp"
+            "zip_file.h")
 
-target_link_libraries(HealthGPS.Datastore PRIVATE HealthGPS.Core fmt::fmt jsoncons)
+target_link_libraries(HealthGPS.Datastore PRIVATE HealthGPS.Core fmt::fmt jsoncons
+                                                  libzippp::libzippp)
 
 set(ROOT_NAMESPACE hgps::data)
--- a/src/HealthGPS.Datastore/CMakeLists.txt
+++ b/src/HealthGPS.Datastore/CMakeLists.txt
@@ -4,7 +4,8 @@ find_package(jsoncons CONFIG REQUIRED)
 add_library(HealthGPS.Datastore STATIC "")
 target_compile_features(HealthGPS.Datastore PUBLIC cxx_std_${CMAKE_CXX_STANDARD})
 
-target_sources(HealthGPS.Datastore PRIVATE "datamanager.cpp" "datamanager.h" "api.h")
+target_sources(HealthGPS.Datastore PRIVATE "api.h" "datamanager.cpp" "datamanager.h" "schema.cpp"
+                                           "schema.h")
 
 target_link_libraries(HealthGPS.Datastore PRIVATE HealthGPS.Core fmt::fmt jsoncons)
 
--- a/src/HealthGPS.Datastore/datamanager.cpp
+++ b/src/HealthGPS.Datastore/datamanager.cpp
@@ -45,13 +45,15 @@ nlohmann::json read_input_files_from_directory(const std::filesystem::path &root
 namespace hgps::data {
 DataManager::DataManager(std::filesystem::path path, VerboseMode verbosity)
     : verbosity_{verbosity} {
-    if (std::filesystem::is_regular_file(path)) {
+    if (std::filesystem::is_directory(path)) {
+        root_ = std::move(path);
+    } else if (std::filesystem::is_regular_file(path) && path.extension() == ".zip") {
         // If it's a file, assume it's a zip file
         root_ = create_temporary_directory();
         extract_zip_file(path, root_);
     } else {
-        // Otherwise it's a folder
-        root_ = std::move(path);
+        throw std::runtime_error(
+            fmt::format("Path must either point to a zip file or a directory: {}", path.string()));
     }
 
     index_ = read_input_files_from_directory(root_);
--- a/src/HealthGPS.Datastore/datamanager.cpp
+++ b/src/HealthGPS.Datastore/datamanager.cpp
@@ -1,8 +1,10 @@
 #include "datamanager.h"
+#include "schema.h"
+#include "zip_file.h"
+
 #include "HealthGPS.Core/math_util.h"
 #include "HealthGPS.Core/string_util.h"
 #include "HealthGPS/program_path.h"
-#include "schema.h"
 
 #include <fmt/color.h>
 #include <rapidcsv.h>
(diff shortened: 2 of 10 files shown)


src/HealthGPS.Datastore/zip_file.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+#include <fmt/format.h>
+#include <libzippp.h>
+
+#include <fstream>
+#include <random>
+
+namespace hgps::data {
+std::filesystem::path create_temporary_directory() {
+    auto tmp_dir = std::filesystem::temp_directory_path();
```
jamesturner246, line 11: I'm wondering how well this behaves on e.g. HPC: is a large array job going to fill up `/tmp` with multiple identical extractions, or does this provide some cleanup in the end?

jamesturner246 (approved):
It looks good. Just one question on how well temp creation behaves on the HPC, and to consider if we want to be a bit careful, or else do some testing.

src/HealthGPS.Datastore/zip_file.cpp
jamesturner246: If it's an issue, perhaps we should extract the zip to the same directory, after checking it doesn't exist or is of bad integrity?

alexdewar: I haven't written any cleanup code myself, but `/tmp` should usually be periodically cleaned by the system so I don't think it will be an issue. In any case, it's only a few megabytes' worth of data that we're extracting anyway (at least for now).

src/HealthGPS.Datastore/CMakeLists.txt
```diff
@@ -1,11 +1,21 @@
...
-target_sources(HealthGPS.Datastore PRIVATE "datamanager.cpp" "datamanager.h" "api.h")
+target_sources(
+    HealthGPS.Datastore
+    PRIVATE "api.h"
+            "datamanager.cpp"
+            "datamanager.h"
+            "schema.cpp"
+            "schema.h"
+            "zip_file.cpp"
+            "zip_file.h")
```
dalonsoa, line 16: For my own benefit could you give me a 1 sentence explanation of what this does?

dalonsoa: A short sentence, obviously :P

src/HealthGPS.Datastore/datamanager.cpp
```diff
@@ -1,86 +1,60 @@
...
     if (version != 2) {
         throw std::runtime_error(fmt::format(
             "File-based store, index schema version: {} mismatch, supported: 2", version));
     }
 
     // Validate against schema
     ifs.seekg(0);
-    const auto schema_directory = get_program_directory() / "schemas" / "v1";
-    validate_index(schema_directory, ifs);
+    const auto schema_directory = hgps::get_program_directory() / "schemas" / "v1";
```
dalonsoa, line 38: This is odd. In the previous line you raise an error if the schema version is not 2, but here you are loading schema v1. Confusing, to say the least. Is this an error?

src/HealthGPS.Datastore/datamanager.cpp
```diff
@@ -1,86 +1,60 @@
...
+
+    return index;
+}
+} // anonymous namespace
+
+namespace hgps::data {
+DataManager::DataManager(std::filesystem::path path, VerboseMode verbosity)
+    : verbosity_{verbosity} {
+    if (std::filesystem::is_regular_file(path)) {
+        // If it's a file, assume it's a zip file
```
dalonsoa, line 49: Cannot you check the extension, at least? It's not foolproof, but it's better than assuming something that might be not - eg. a CSV file.

src/HealthGPS.Datastore/datamanager.cpp
```diff
@@ -1,86 +1,60 @@
...
+        root_ = create_temporary_directory();
+        extract_zip_file(path, root_);
+    } else {
+        // Otherwise it's a folder
```
dalonsoa, line 53: The same, cannot you check if this is a directory rather than assuming anything?

I would include the following branches in the following order:
- Is it a directory?
- Is it a ZIP file?
- If neither, there's an error.

src/HealthGPS.Datastore/zip_file.cpp
jamesturner246: Okay. Maybe we should open an issue to remind ourselves to add cleanup in future, if it becomes an issue.

alexdewar: Yeah, I was a bit worried about the clobbering thing, which is why I wanted to use a unique folder for each extraction.

I'm planning on adding caching for downloaded zip files at some point, so could also use the cache folder for extracing them to? On Linux this would be something like `.cache/healthgps`. We'd have to be a little careful to make sure there aren't race conditions with extracting and reading the zip files (e.g. for array jobs), but this should be doable. What do you think?

src/HealthGPS.Datastore/zip_file.cpp
```diff
@@ -0,0 +1,55 @@
...
+            if (!std::filesystem::create_directories(out_path)) {
+                throw std::runtime_error{
+                    fmt::format("Failed to create directory: {}", out_path.string())};
+            }
+        } else {
+            std::ofstream ofs{out_path};
+            if (!ofs) {
+                throw std::runtime_error{
+                    fmt::format("Failed to create file: {}", out_path.string())};
+            }
```
dalonsoa, line 49: Would it be useful to attach the original error to get an idea of why these fail, so the user can investigate?

dalonsoa (changes requested):
I've a few suggestions, mainly related to error handling and explicitly. Maybe I'm in a pythonic mindset, but I thought worth to highlight them.

src/HealthGPS.Datastore/CMakeLists.txt
alexdewar: It's just a list of source files for the `HealthGPS.Datastore` library. You have to be explicit with CMake/C++.

(Sorry that was two sentences :stuck_out_tongue:.)

src/HealthGPS.Datastore/datamanager.cpp
alexdewar: It's not. This data file has had a `version` property for ages now (and I think some of the other input files have one too) but the problem is that it doesn't seem like they were particularly careful with backwards compatibility, so the field is basically meaningless anyway.

I was going to remove this property because it's pointless now that we're using schemas. Shall I open an issue for it?

dalonsoa: Indeed!

src/HealthGPS.Datastore/zip_file.cpp
alexdewar: The only reason it'll fail here is because of an IO error and C++ won't give any information beyond that, unfortunately.

src/HealthGPS.Datastore/zip_file.cpp
jamesturner246: Sounds good. :+1:  So long as those race conditions can be avoided!

src/HealthGPS.Datastore/datamanager.cpp
alexdewar: https://github.com/imperialCHEPI/healthgps/issues/399

src/HealthGPS.Datastore/zip_file.cpp
alexdewar: I think the trick is to create a temporary folder with a unique name in the same folder and then rename it once the extraction is complete.

alexdewar: #400

src/HealthGPS.Datastore/datamanager.cpp
alexdewar: Done

dalonsoa (approved):

PR #393, opened 2024-05-29
`pre-commit`: Add hook to check metaschemas

alexdewar:
Add a `pre-commit` hook to check that JSON schemas are correct.

Sorry for this piddling little PR, but I thought this should be submitted by itself.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,6 +11,8 @@ repos:
     rev: 0.28.4
     hooks:
       - id: check-github-workflows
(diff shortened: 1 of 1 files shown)


jamesturner246 (approved):
:+1: 

How does validation work?

alexdewar:
It only validates the actual schemas themselves are correctly formed (i.e. the files in `schemas/`). There's actually a schema for validating JSON schemas so it can use that.

We probably also want to verify that our input files themselves work with *our* schemas at some stage.

PR #402, opened 2024-05-30
Remove pointless `version` property from data `index.json` files

### LINKED ISSUE #399: Remove pointless `version` property from data `index.json` files
It's not. This data file has had a `version` property for ages now (and I think some of the other input files have one too) but the problem is that it doesn't seem like they were particularly careful with backwards compatibility, so the field is basically meaningless anyway.

I was going to remove this property because it's pointless now that we're using schemas. Shall I open an issue for it?

_Originally posted by @alexdewar in https://github.com/imperialCHEPI/healthgps/pull/391#discussion_r1620549520_

alexdewar:
This PR removes the pointless `version` property, as discussed.

I've also opted to add a `$schema` property in its place which is the standard way to indicate which schema should be used for a JSON document.

Closes #399.

### DIFF
--- a/data/index.json
+++ b/data/index.json
@@ -1,4 +1,5 @@
 {
+    "$schema": "https://raw.githubusercontent.com/imperialCHEPI/healthgps/main/schemas/v1/data_index.json",
     "version": 2,
     "country": {
         "format": "csv",
--- a/data/index.json
+++ b/data/index.json
@@ -1,6 +1,5 @@
 {
     "$schema": "https://raw.githubusercontent.com/imperialCHEPI/healthgps/main/schemas/v1/data_index.json",
-    "version": 2,
     "country": {
         "format": "csv",
         "delimiter": ",",
--- a/schemas/v1/data_index.json
+++ b/schemas/v1/data_index.json
@@ -5,6 +5,9 @@
     "description": "Metadata and paths for Health-GPS input files",
     "type": "object",
     "properties": {
+        "$schema": {
+            "const": "https://raw.githubusercontent.com/imperialCHEPI/healthgps/main/schemas/v1/data_index.json"
+        },
         "version": { "const": 2 },
         "country": {
             "allOf": [
--- a/schemas/v1/data_index.json
+++ b/schemas/v1/data_index.json
@@ -8,7 +8,6 @@
         "$schema": {
             "const": "https://raw.githubusercontent.com/imperialCHEPI/healthgps/main/schemas/v1/data_index.json"
         },
-        "version": { "const": 2 },
         "country": {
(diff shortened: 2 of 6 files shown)


data/index.json
```diff
@@ -1,5 +1,5 @@
 {
-    "version": 2,
+    "$schema": "https://raw.githubusercontent.com/imperialCHEPI/healthgps/main/schemas/v1/data_index.json",
```
dalonsoa, line 2: What if the user does not include this?

schemas/v1/data_index.json
```diff
@@ -5,6 +5,9 @@
     "description": "Metadata and paths for Health-GPS input files",
     "type": "object",
     "properties": {
+        "$schema": {
+            "const": "https://raw.githubusercontent.com/imperialCHEPI/healthgps/main/schemas/v1/data_index.json"
+        },
         "version": { "const": 2 },
```
dalonsoa, line 11: Shall this be removed?

src/HealthGPS.Datastore/datamanager.cpp
```diff
@@ -21,18 +21,9 @@ nlohmann::json read_input_files_from_directory(const std::filesystem::path &root
...
-
-    auto version = index["version"].get<int>();
-    if (version != 2) {
-        throw std::runtime_error(fmt::format(
-            "File-based store, index schema version: {} mismatch, supported: 2", version));
-    }
-
     // Validate against schema
     ifs.seekg(0);
     const auto schema_directory = hgps::get_program_directory() / "schemas" / "v1";
```
dalonsoa, line 29: If the information on the schema to use is included in the `index.json` file, shall you use that instead?

dalonsoa (approved):
The change is ok, but don't you need to remove some extra stuff no longer needed?

schemas/v1/data_index.json
alexdewar: Yes! Oops :see_no_evil:.

What's slightly concerning is that the file is still validating... I think this must mean "if the version field is present, it must be 2", but it doesn't enforce its presence. Ugh.

I've opened an issue for it: #403

src/HealthGPS.Datastore/datamanager.cpp
alexdewar: I don't actually want users to be able to use different schemas. My current plan is to enforce that they must use the latest schema with the latest version of Health-GPS, but at some point we might want to revisit this for the sake of backwards compatibility.

jamesturner246 (approved):
Looks good. Go ahead.

alexdewar:
Thanks!

PR #404, opened 2024-05-30
Make properties required for JSON schema

### LINKED ISSUE #403: Enforce presence of some/all fields in data `index.json` file
I've released that the schema for the `index.json` file isn't quite right as it stands.

There are two problems:

1. Users can omit any of the specified properties and the file will still validate
2. Users can add arbitrary additional properties

The second is less serious (though still potentially a source of user error), but the
first is important.

The solutions are:

1. Making use of the `required` keyword throughout the schema
2. Adding `additionalProperties: false` throughout

Yuck.

alexdewar:
For `file_info.json` I made only the `path` property required, because we don't actually use the others anywhere in the code (though they may be useful metadata).

In the process, I discovered that one field was in the wrong section (it's hard to keep track of all those curly brackets!), so I've fixed that too.

I used the following script to make all properties for `data_index.json` required:

```py
import json
import sys


def add_required(d: dict):
    required = []
    for key, value in d.items():
        if isinstance(value, dict):
            add_required(value)
        if isinstance(value, list):
            for v in value:
                if isinstance(v, dict):
                    add_required(v)
        if key == "properties":
            required = list(k for k in d['properties'].keys() if k != "$schema")
    
    if required:
        d['required'] = required
        # d['additionalProperties'] = False


def process_file(path):
    with open(path) as f:
        d = json.load(f)
    
    add_required(d)

    with open(path, 'w') as f:
        json.dump(d, f)


if __name__ == '__main__':
    for path in sys.argv[1:]:
        process_file(path)
```

Making `additionalProperties` `false` seemed to break validation for some reason, so I've omitted it for now because it's less important.

Closes #403.

### DIFF
--- a/schemas/v1/data_index.json
+++ b/schemas/v1/data_index.json
@@ -12,9 +12,8 @@
                 },
                 {
                     "type": "object",
-                    "properties": {
-                        "file_name": { "type": "string" }
-                    }
+                    "properties": { "file_name": { "type": "string" } },
+                    "required": ["file_name"]
                 }
             ]
         },
@@ -51,7 +50,8 @@
                                 "description": { "type": "string" },
                                 "path": { "type": "string" },
                                 "file_name": { "type": "string" }
-                            }
+                            },
+                            "required": ["description", "path", "file_name"]
                         },
                         "mortality": {
                             "type": "object",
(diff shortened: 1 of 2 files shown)


jamesturner246 (changes requested):
Presumably you want `"maxItems": 2,` for the range pairs as well? Also a few of the dicts still don't have `required`, e.g. `prefixItems`.

Also what does `additionalProperties=false` do?

jamesturner246:
Okay. But still `maxItems` for pairs though? Or no?

alexdewar:
Sorry, I misread that! I thought you said `minItems`... I'll fix that

alexdewar:
I've just looked at this and it does actually work as is, i.e. if you supply 3 values instead of 2 you get an error. The reason is that the `prefixItems` array enforces that you can't have more than two items (as well as making sure they're both integers).

It seems JSON schemas are not that intuitive :stuck_out_tongue:

jamesturner246 (approved):
LGTM!

Issue #376, opened 2024-05-20
Add ability to group by discrete variables, instead of just age and gender (e.g. income, age, gender)


TinyMarsh:
- [x] #419
- [ ] #420

PR #405, opened 2024-05-30
Group by discrete variables

TinyMarsh:
This is by no means a PR with code suggestions. I just wanted to use the diff to facilitate discussion and to get help.

Context: #376. We need to add the ability to group by discrete variables, instead of just age and gender. So if I wanted to group by age and `over_18`, I could change the variable `gender` to `over_18` and fetch the correct value from the `Person` object, i.e. `entity.over_18`.

Obviously this isn't going to work as is, for various reasons. Namely that `series` needs to be called with a gender, not a bool. And that we need to be able to group by other discrete variables in _addition_ to age/gender, not _instead_ of. But is this at least on the right track or am I barking up the wrong tree?


alexdewar:
Honestly, I don't know this part of the code at all so I have no idea! I'm happy to dig into it if needed though... @jamesturner246 how well do you know it?

PR #415, opened 2024-06-04
Colourise compiler output

alexdewar:
Normally, the fact that the compiler output is piped through cmake means that colours are turned off by default. Explicitly enable them for the sake of readability.

Handy for spotting warnings/errors on your local machine.


jamesturner246 (approved):
LGTM!

PR #416, opened 2024-06-04
CI: Reinstate dependabot's permissions

### LINKED ISSUE #409: [BUG] Dependabot PRs all failing CI checks
Specifically, the clang-tidy workflow seems to be failing to post messages to PRs. See e.g.: https://github.com/imperialCHEPI/healthgps/actions/runs/9300334072/job/25726763314

This seems spurious rather than anything that dependabot is doing wrong.

alexdewar:
It seems that dependabot no longer has write permissions for GitHub tokens by default. This can be worked around by changing the "on" field from "pull_request" to "pull_request_target".

NB: I haven't actually been able to test this, but according to the upstream developers, this is how you can fix the problem.

Fixes #409.

See: https://github.com/dependabot/dependabot-core/issues/3253

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,7 @@ on:
     branches: [main]
   release:
     types: [published]
-  pull_request:
+  pull_request_target:
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,6 +38,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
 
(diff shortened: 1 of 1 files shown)


TinyMarsh:
Interesting. A bit more discussion about security of `pull_request_target` here: https://nathandavison.com/blog/github-actions-and-the-threat-of-malicious-pull-requests

TinyMarsh (approved):

alexdewar:
Thanks for the link @TinyMarsh. It seems like what I am trying to do is actually the insecure way :stuck_out_tongue:. Also the tests don't seem to be running for this PR... maybe GitHub knows I'm trying to do something daft? Dunno. I guess this will take a bit more thought.

alexdewar:
Superseded by #417

PR #417, opened 2024-06-04
CI: Don't run clang-tidy checks for dependabot PRs

### LINKED ISSUE #409: [BUG] Dependabot PRs all failing CI checks

### LINKED ISSUE #410: Auto-merge PRs from bots
Given that the bot PRs are only for updating pre-commit hooks and GitHub Actions, there's probably no risk attached to this.

We could use something like the config here: https://github.com/ImperialCollegeLondon/pip-tools-template/blob/main/.github/workflows/auto-merge.yml

alexdewar:
Ok, here's my less horribly insecure attempt at fixing this. If we just don't run the `clang-tidy` checks for PRs from bots, that should fix things.

While I was at it, I borrowed Adrian's workflow for auto-merging bot PRs, because I thought that would be useful.

Fixes #409. Closes #410.

### DIFF
--- a/.github/workflows/auto-merge.yml
+++ b/.github/workflows/auto-merge.yml
@@ -1,0 +1,21 @@
+name: Dependabot and Pre-commit auto-merge
+
+on:
+  pull_request:
+
+permissions:
+  contents: write
+  pull-requests: write # Needed if in a private repository
+
+jobs:
+  auto-merge:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


jamesturner246 (approved):
LGTM. :+1:

Issue #326, opened 2024-02-28
Compute standard deviation in output files


jamesturner246:
Standard deviation of factors for all age/sex group should also be computed and added to the output CSV file.

jackolney:
Is this still active?

jamesturner246:
Yes -- see #430.

jzhu20:
Results generated with #430 


![Image](https://github.com/imperialCHEPI/healthgps/assets/154426800/dd759820-88c0-46f3-ac30-55d9ad396ad9)



![Image](https://github.com/imperialCHEPI/healthgps/assets/154426800/f10b1b41-ed98-4a84-8d4e-03c6cde4173a)



![Image](https://github.com/imperialCHEPI/healthgps/assets/154426800/1ed02b7a-cc9f-4763-92f3-689db2611f1f)

PR #430, opened 2024-06-06
Compute standard deviation in output.

jamesturner246:
Addresses #326. 

In this change, the standard deviation for each risk factor is computed in the `std_<factor>` columns of the CSV output.

Group means are now in the `mean_<factor>` columns. NB: this probably means scripts need changing to point to the new `mean_` prefixed equivalents.

Note that *ONLY* risk factors have stddev computed for each groups. If more columns are required to have stddev, please let me know here.

@jzhu20, could you please test locally so it works your end if/when you have a moment? Thanks.

### DIFF
--- a/src/HealthGPS/analysis_module.cpp
+++ b/src/HealthGPS/analysis_module.cpp
@@ -348,9 +348,12 @@ void AnalysisModule::calculate_standard_deviation(RuntimeContext &context, DataS
     int max_age = context.age_range().upper();
 
     for (const auto &person : context.population()) {
+        if (!person.is_active()) {
+            continue;
+        }
+
         unsigned int age = person.age;
         core::Gender sex = person.gender;
-
         for (const auto &factor : context.mapping().entries()) {
             double value = person.get_risk_factor_value(factor.key());
             double mean = series(sex, "mean_" + factor.key().to_string()).at(age);
--- a/src/HealthGPS/analysis_module.cpp
+++ b/src/HealthGPS/analysis_module.cpp
@@ -343,8 +343,7 @@ void AnalysisModule::calculate_population_statistics(RuntimeContext &context,
 }
 // NOLINTEND(readability-function-cognitive-complexity)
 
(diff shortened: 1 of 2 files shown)


src/HealthGPS/analysis_module.cpp
```diff
@@ -326,28 +326,64 @@ void AnalysisModule::calculate_population_statistics(RuntimeContext &context,
     // Calculate in-place averages
     for (auto i = min_age; i <= max_age; i++) {
```
alexdewar, line 327: I know this is out of scope for this PR, but this could be better vectorised. Not sure if performance is an issue here though...

alexdewar (approved):
LGTM, provided it works :stuck_out_tongue:

src/HealthGPS/analysis_module.cpp
jamesturner246: Could just add TBB parallel_for in all the outer loops right now.

src/HealthGPS/analysis_module.cpp
```diff
@@ -326,28 +326,64 @@ void AnalysisModule::calculate_population_statistics(RuntimeContext &context,
...
         for (const auto &chan : series.channels()) {
-            if (chan == "count") {
+            if (chan == "count" || chan.starts_with("std_")) {
```
TinyMarsh, line 329: Not really a suggestion, but why isn't this just
```suggestion
            if (!chan.starts_with("mean_")) {
```

TinyMarsh (approved):
LGTM

src/HealthGPS/analysis_module.cpp
jamesturner246: Ah, right, so some of the columns are binary variables, which are normalised to fractions using the division below. Like the disease `incidence_` and `prevalence_` fields. Maybe a better comment/name for this would be computing means *and* fractions.

alexdewar:
@jzhu20 This PR is already merged, so I'd open a new issue for this request.

PR #431, opened 2024-06-07
Initialise calculated statistics map

### LINKED ISSUE #419: Add data structure, and code for clearing, for storing analysis means
For now add a simple (2D?) private data structure (`unordered_map`?) To the analysis class.

Add code to the analysis `update_population` to reset values in this structure to zero.

We can later worry about choosing the correct data structure and refining it.

TinyMarsh:
Adds an unordered_map as a private data member to `AnalysisModule` class called `calculated_factors_`. This is where the calculated population statistics will live. Also adds code to `update_population` to zero the inner map values.

Closes #419

### DIFF
--- a/src/HealthGPS/analysis_module.cpp
+++ b/src/HealthGPS/analysis_module.cpp
@@ -55,8 +55,7 @@ void AnalysisModule::initialise_vector(RuntimeContext &context) {
 
     // The product of the number of bins for each factor can be used to calculate the size of the
     // `calculated_factors_` in the next step
-    auto total_num_bins =
-        std::accumulate(factor_bins.cbegin(), factor_bins.cend(), 1, std::multiplies<int>());
+    auto total_num_bins = std::ranges::accumulate(factor_bins, 1, std::multiplies<int>());
 
     // Set the vector size and initialise all values to 0.0
     calculated_factors_.resize(total_num_bins * num_factors_to_calc);
--- a/src/HealthGPS/analysis_module.cpp
+++ b/src/HealthGPS/analysis_module.cpp
@@ -26,6 +26,44 @@ SimulationModuleType AnalysisModule::type() const noexcept {
     return SimulationModuleType::Analysis;
 }
 
+void AnalysisModule::initialise_map(RuntimeContext &context) {
+    // Set some arbitrary factors
+    auto factor1 = context.mapping().entries().at(0).key().to_string();
+    auto factor2 = context.mapping().entries().at(1).key().to_string();
+
+    // Get the minimum and maximum of the two factors in the population
+    auto min_max = std::minmax_element(context.population().cbegin(), context.population().cend(),
+                                       [](const auto &a, const auto &b) {
+                                           return a.get_risk_factor_value(factor1) <
+                                                  b.get_risk_factor_value(factor1);
+                                       });
+
+    auto min_factor1 = min_max.first->get_risk_factor_value(factor1);
+    auto max_factor1 = min_max.second->get_risk_factor_value(factor1);
+
+    min_max = std::minmax_element(context.population().cbegin(), context.population().cend(),
+                                  [](const auto &a, const auto &b) {
+                                      return a.get_risk_factor_value(factor2) <
+                                             b.get_risk_factor_value(factor2);
+                                  });
+
+    auto min_factor2 = min_max.first->get_risk_factor_value(factor2);
+    auto max_factor2 = min_max.second->get_risk_factor_value(factor2);
+
+    // The dimensions of the 2D map is the number of integer values of each factor, or 100 bins of equal size, whichever is smaller
+    // (100 is an arbitrary number, it could be any other number depending on the desired resolution of the map)
+    auto factor1_bins = std::min(100, static_cast<int>(max_factor1 - min_factor1));
+    auto factor2_bins = std::min(100, static_cast<int>(max_factor2 - min_factor2));
+
+    // Create a 2D map with the specified number of bins
+    calculated_factors_ = std::map<int, std::map<int, std::map<std::string, double>>>{};
+    for (int i = 0; i < factor1_bins; i++) {
+        for (int j = 0; j < factor2_bins; j++) {
+            calculated_factors_[i][j] = std::map<std::string, double>{};
+        }
+    }
+}
+
 const std::string &AnalysisModule::name() const noexcept { return name_; }
 
 void AnalysisModule::initialise_population(RuntimeContext &context) {
@@ -66,7 +104,27 @@ void AnalysisModule::initialise_population(RuntimeContext &context) {
     publish_result_message(context);
 }
 
-void AnalysisModule::update_population(RuntimeContext &context) { publish_result_message(context); }
+void AnalysisModule::update_population(RuntimeContext &context) {
+
+    // The inner map will have keys which are every factor except the two we are analysing
+    // The value of each key will be reset to 0.0
+    std::map<std::string, double> inner_map{};
+    for (const auto &entry : context.mapping().entries()) {
+        if (entry.key() != context.mapping().entries().at(0).key() &&
+            entry.key() != context.mapping().entries().at(1).key()) {
+            inner_map[entry.key().to_string()] = 0.0;
+        }
+    }
+
+    // Iterate over every element in calculated_factors_ and set its value to inner map
+    for (auto &outer : calculated_factors_) {
+        for (auto &inner : outer.second) {
+            inner.second = inner_map;
+        }
+    }
+
+    publish_result_message(context);
+}
 
 double
 AnalysisModule::calculate_residual_disability_weight(int age, const core::Gender gender,
--- a/src/HealthGPS/analysis_module.cpp
+++ b/src/HealthGPS/analysis_module.cpp
@@ -34,9 +34,11 @@ void AnalysisModule::initialise_vector(RuntimeContext &context) {
     factor_bins.reserve(factors.size());
 
     for (const auto &factor : factors) {
-        auto min_max = std::ranges::minmax_element(context.population(), [&factor](const auto &entity1, const auto &entity2) {
-            return entity1.get_risk_factor_value(factor) < entity2.get_risk_factor_value(factor);
-        });
+        auto min_max = std::ranges::minmax_element(
+            context.population(), [&factor](const auto &entity1, const auto &entity2) {
+                return entity1.get_risk_factor_value(factor) <
+                       entity2.get_risk_factor_value(factor);
+            });
 
         auto min_factor = min_max.first->get_risk_factor_value(factor);
         auto max_factor = min_max.second->get_risk_factor_value(factor);
--- a/src/HealthGPS/analysis_module.cpp
+++ b/src/HealthGPS/analysis_module.cpp
@@ -55,8 +55,8 @@ void AnalysisModule::initialise_vector(RuntimeContext &context) {
 
     // The product of the number of bins for each factor can be used to calculate the size of the
     // `calculated_factors_` in the next step
-    auto total_num_bins = std::accumulate(factor_bins.cbegin(), factor_bins.cend(), 1,
-                                    std::multiplies<>());
+    auto total_num_bins =
+        std::accumulate(factor_bins.cbegin(), factor_bins.cend(), 1, std::multiplies<>());
 
     // Set the vector size and initialise all values to 0.0
     calculated_factors_.resize(total_num_bins * num_factors_to_calc);
--- a/src/HealthGPS/analysis_module.cpp
+++ b/src/HealthGPS/analysis_module.cpp
@@ -55,8 +55,8 @@ void AnalysisModule::initialise_vector(RuntimeContext &context) {
 
     // The product of the number of bins for each factor can be used to calculate the size of the
     // `calculated_factors_` in the next step
-    auto total_num_bins = std::accumulate(factor_bins.cbegin(), factor_bins.cend(), 1,
-                                    std::multiplies<int>());
+    auto total_num_bins =
+        std::accumulate(factor_bins.cbegin(), factor_bins.cend(), 1, std::multiplies<int>());
 
     // Set the vector size and initialise all values to 0.0
     calculated_factors_.resize(total_num_bins * num_factors_to_calc);
--- a/src/HealthGPS/analysis_module.cpp
+++ b/src/HealthGPS/analysis_module.cpp
@@ -50,8 +50,9 @@ void AnalysisModule::initialise_map(RuntimeContext &context) {
     auto min_factor2 = min_max.first->get_risk_factor_value(factor2);
     auto max_factor2 = min_max.second->get_risk_factor_value(factor2);
 
-    // The dimensions of the 2D map is the number of integer values of each factor, or 100 bins of equal size, whichever is smaller
-    // (100 is an arbitrary number, it could be any other number depending on the desired resolution of the map)
+    // The dimensions of the 2D map is the number of integer values of each factor, or 100 bins of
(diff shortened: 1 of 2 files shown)


alexdewar:
@TinyMarsh Re these `clang-tidy` warnings, I'm working on it: #432

jamesturner246 (changes requested):
Does the job nicely. Just a few points to get it nailed down.

src/HealthGPS/analysis_module.cpp
```diff
@@ -26,6 +26,46 @@ SimulationModuleType AnalysisModule::type() const noexcept {
...
 }
 
+void AnalysisModule::initialise_map(RuntimeContext &context) {
+    // Set some arbitrary factors
+    auto factor1 = context.mapping().entries().at(0).key().to_string();
+    auto factor2 = context.mapping().entries().at(1).key().to_string();
+
+    // Get the minimum and maximum of the two factors in the population
+    auto min_max = std::minmax_element(context.population().cbegin(), context.population().cend(),
+                                       [](const auto &a, const auto &b) {
```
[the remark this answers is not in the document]
alexdewar: The linter warning about the names is one that should be turned off... If you merge in `main` you'll get my fix for that and it should stop complaining

src/HealthGPS/analysis_module.cpp
```diff
@@ -66,7 +106,27 @@ void AnalysisModule::initialise_population(RuntimeContext &context) {
     publish_result_message(context);
 }
 
-void AnalysisModule::update_population(RuntimeContext &context) { publish_result_message(context); }
+void AnalysisModule::update_population(RuntimeContext &context) {
+
+    // The inner map will have keys which are every factor except the two we are analysing
```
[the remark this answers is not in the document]
alexdewar: Instead of `unused`, you can write `std::ignore`

src/HealthGPS/analysis_module.cpp
```diff
@@ -26,6 +26,46 @@ SimulationModuleType AnalysisModule::type() const noexcept {
     return SimulationModuleType::Analysis;
 }
 
+void AnalysisModule::initialise_map(RuntimeContext &context) {
+    // Set some arbitrary factors
+    auto factor1 = context.mapping().entries().at(0).key().to_string();
+    auto factor2 = context.mapping().entries().at(1).key().to_string();
+
+    // Get the minimum and maximum of the two factors in the population
+    auto min_max = std::minmax_element(context.population().cbegin(), context.population().cend(),
```
alexdewar, line 35: While this is totally fine as it is, in C++20 you *can* write this as:

```suggestion
    auto min_max = std::ranges::minmax_element(context.population(),
```

src/HealthGPS/analysis_module.cpp
```diff
@@ -26,6 +26,46 @@ SimulationModuleType AnalysisModule::type() const noexcept {
...
+                                       [](const auto &a, const auto &b) {
+                                           return a.get_risk_factor_value(factor1) <
+                                                  b.get_risk_factor_value(factor1);
+                                       });
+
+    auto min_factor1 = min_max.first->get_risk_factor_value(factor1);
+    auto max_factor1 = min_max.second->get_risk_factor_value(factor1);
+
+    min_max = std::minmax_element(context.population().cbegin(), context.population().cend(),
```
alexdewar, line 44: See above

src/HealthGPS/analysis_module.cpp
```diff
@@ -26,6 +26,46 @@ SimulationModuleType AnalysisModule::type() const noexcept {
...
+
+    // The dimensions of the 2D map is the number of integer values of each factor, or 100 bins of
+    // equal size, whichever is smaller (100 is an arbitrary number, it could be any other number
+    // depending on the desired resolution of the map)
+    auto factor1_bins = std::min(100, static_cast<int>(max_factor1 - min_factor1));
+    auto factor2_bins = std::min(100, static_cast<int>(max_factor2 - min_factor2));
+
+    // Create a 2D map with the specified number of bins
+    calculated_factors_ =
+        std::unordered_map<int, std::unordered_map<int, std::unordered_map<std::string, double>>>{};
```
alexdewar, line 61: I'm assuming that this method will only be called once on class creation, right? In that case, `calculated_factors_` will already be initialised as an empty map, so you don't need to do it again:

```suggestion
```

If you *do* need to reinitialise it, you can call `calculated_factors_.clear()` instead.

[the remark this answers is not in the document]
alexdewar: I'm guessing the keys will be in some range 0 to `n` for the outer map and 0 to `m` for the inner one? If that's the case, then you don't need maps; you could just use `std::vector`s instead, which will be faster.

I.e. the type would be `std::vector<std::vector<std::unordered_map<core::Identifier, double>>>`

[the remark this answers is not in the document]
alexdewar: (Technically you could just use a single `std::vector` and then access the right element with `i * row_size + j` which will be faster still, but I don't know whether we care that much about performance here. It's also *potentially* a bit cleaner, because it makes it easier to do things to the whole data structure at once.)

src/HealthGPS/analysis_module.cpp
```diff
@@ -26,6 +26,46 @@ SimulationModuleType AnalysisModule::type() const noexcept {
...
+    for (int i = 0; i < factor1_bins; i++) {
+        for (int j = 0; j < factor2_bins; j++) {
+            calculated_factors_[i][j] = std::unordered_map<std::string, double>{};
```
alexdewar, line 64:
```suggestion
            calculated_factors_[i][j].emplace();
```

That said, if you follow my suggestion above of using one or two `std::vector`s, you just need to resize them and the contents will be default-initialised.

src/HealthGPS/analysis_module.cpp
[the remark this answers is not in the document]
alexdewar: See above: use `std::vector`s and then it's easier to reset the contents

src/HealthGPS/analysis_module.cpp
```diff
@@ -66,7 +106,27 @@ void AnalysisModule::initialise_population(RuntimeContext &context) {
...
+            inner_map[entry.key().to_string()] = 0.0;
+        }
+    }
+
+    // Iterate over every element in calculated_factors_ and set its value to inner map
+    for (auto &outer : calculated_factors_) {
+        for (auto &inner : outer.second) {
+            inner.second = inner_map;
+        }
+    }
```
alexdewar, line 126: Given that this is just a default value, if using `std::vector`(s) you can use `std::fill` to reset the conents.

alexdewar (changes requested):
I may have misunderstood something here, but I think `calculated_factors_` could be a vector instead of a map (or a vector of vectors), which will make your life easier too

alexdewar:
Ok, final suggestion re the data structure! I've realised it could just be a 2D matrix (Eigen::MatrixXd), with a separate vector of strings representing the column headings. I think that's probably the cleanest/easiest way to do things

TinyMarsh:
Hmm, I'm not sure how I can replace the current data structure with Eigen::MatrixXd. Since Eigen::MatrixXd only deals with numerical types, I can't have the values of the Eigen::MatrixXd as `unordered_map`s.

The keys of the outer two umaps are 0 to `m` and 0 to `n`, as @alexdewar says, but the keys of the innermost map are `core::Identifier` (as @jamesturner246 suggested).

Maybe I should just sack off this horrible nested umap stuff and store all the values in one `std::vector` and access the elements with `i * row_size + j * col_size + k`, similar to how @alexdewar suggested above. That way everything is in a `std::vector`, which is nicer and easier to deal with?

alexdewar:
@TinyMarsh Yep, I think a single `std::vector` would work.

I was assuming that all of the inner `std::unordered_map`s (i.e. the ones with a string/Identifier key) had the same keys (i.e. that they're column names). If so, you could store the column names separately and then you'd just have a normal matrix to work with for the numbers; when you need the column names you could fish them out of this other data structure based on the column number. But if the sets of keys are all different then this won't work at all!

jamesturner246:
If we are thinking about the final data structure, then one more option would be to just go straight for the [Apache Arrow `Tensor` class](https://arrow.apache.org/docs/cpp/api/tensor.html). That's the direction we are heading in anyway, right?

I think the backend buffer is still a contiguous memory block anyway, so single `std::vector` is fine either way if you don't fancy using `Tensor`s yet.

jamesturner246:
EDIT: 
We could then easily flip over to the Arrow `Table` class, with its powerful group-by methods, once `Population` data is contiguous.It could be like:

```
| key_1 | ... | key_ | col_1 | ... col_n |
------------------------------------------------
| key_1_val_1 | ... | key_n_val_1 | cols...
| key_1_val_1 | ... | key_n_val_2 | cols...
| key_1_val_2 | ... | key_n_val_1 | cols...
| key_1_val_2 | ... | key_n_val_2 | cols...
```

You'd still need to have the `core::Identifier` vectors of key names and column names, but this should be flexible enough to use multiple keys.

alexdewar:
@jamesturner246 Potentially. You can also do tensors with Eigen though.

I did have a brief look at Apache Arrow and the API for C++ seemed kinda gross :disappointed:

jamesturner246:
I had a look at Eigen tensors. Noticed it was an 'unsupported' extension. [see here](https://eigen.tuxfamily.org/dox/unsupported/eigen_tensors.html). Which one are you looking at?

One thing to keep in mind is that we eventually want to group by an arbitrary number of keys

src/HealthGPS/analysis_module.cpp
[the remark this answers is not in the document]
jamesturner246: I think we had this come up in a prior PR, where it turned out you can't use `std::ignore` in a structured binding -- least not in C++20. Hence the above. I can't remember which PR.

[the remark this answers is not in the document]
alexdewar: Ah, I see. Nvm then!

src/HealthGPS/analysis_module.cpp
```diff
@@ -26,6 +26,41 @@ SimulationModuleType AnalysisModule::type() const noexcept {
     return SimulationModuleType::Analysis;
 }
 
+void AnalysisModule::initialise_vector(RuntimeContext &context) {
+    // Set some arbitrary factors
+    std::vector<core::Identifier> factors{"Gender"_id, "Age"_id};
+
+    std::vector<int> factor_bins;
```
jamesturner246, line 33: To avoid potential reallocations (unlikely with few keys, but still), one can `reserve` space inside the vector before pushing back.
```suggestion
    std::vector<int> factor_bins;
    factor_bins.reserve(factors.size());
```

src/HealthGPS/analysis_module.cpp
```diff
@@ -66,7 +101,13 @@ void AnalysisModule::initialise_population(RuntimeContext &context) {
     publish_result_message(context);
 }
 
-void AnalysisModule::update_population(RuntimeContext &context) { publish_result_message(context); }
+void AnalysisModule::update_population(RuntimeContext &context) {
+
+    // Reset the calculated factors vector to 0.0
+    std::fill(calculated_factors_.begin(), calculated_factors_.end(), 0.0);
```
jamesturner246, line 107: Might be able to use `std::ranges::fill(calculated_factors_, 0.0)` here, but it *might* be C++26. Also insertion defaults to `0.0` anyway, so value is not needed.

src/HealthGPS/analysis_module.cpp
```diff
@@ -26,6 +26,41 @@ SimulationModuleType AnalysisModule::type() const noexcept {
...
+    std::vector<core::Identifier> factors{"Gender"_id, "Age"_id};
+
+    std::vector<int> factor_bins;
+
+    for (const auto &factor : factors) {
+        auto min_max = std::minmax_element(
+            context.population().cbegin(), context.population().cend(),
+            [&factor](const auto &a, const auto &b) {
+                return a.get_risk_factor_value(factor) < b.get_risk_factor_value(factor);
+            });
```
jamesturner246, line 40: Use the [std::ranges version](https://en.cppreference.com/w/cpp/algorithm/ranges/minmax_element):

`std::minmax_element(context.population(), ...`

jamesturner246 (changes requested):
Looks good. Some tiny extra things to polish off.

src/HealthGPS/analysis_module.cpp
```diff
@@ -26,6 +26,41 @@ SimulationModuleType AnalysisModule::type() const noexcept {
...
+        // any other number depending on the desired resolution of the map)
+        factor_bins.push_back(std::min(100, static_cast<int>(max_factor - min_factor)));
+    }
+
+    // The number of factors to calculate is the number of factors minus the length of the `factors`
+    // vector.
+    auto num_factors_to_calc = context.mapping().entries().size() - factors.size();
+
+    // The product of the number of bins for each factor can be used to calculate the size of the
+    // `calculated_factors_` in the next step
```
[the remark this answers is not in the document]
jamesturner246: Is this fixed now?

src/HealthGPS/analysis_module.cpp
alexdewar: Sometimes `clang-tidy` will give this advice too, but it seems not to have done here for some reason

src/HealthGPS/analysis_module.cpp
alexdewar: I'm assuming this is placeholder code and that something else will follow resetting the elements to zero? Fine for now, but either way @jamesturner246 is right that eventually you might not need to explicitly zero the vector

src/HealthGPS/analysis_module.cpp
```diff
@@ -26,44 +26,38 @@ SimulationModuleType AnalysisModule::type() const noexcept {
...
+        // depending on the desired resolution of the map)
+        factor_bins.push_back(std::min(100, static_cast<int>(max_factor - min_factor)));
     }
+
+    // The number of factors to calculate is the number of factors minus the length of the `factors` vector.
+    auto num_factors_to_calc = context.mapping().entries().size() - factors.size();
+
+    // The product of the number of bins for each factor can be used to calculate the size of the `calculated_factors_` in the next step
+    auto total_num_bins = std::accumulate(factor_bins.cbegin(), factor_bins.cend(), 1,
+                                    std::multiplies<int>());
```
alexdewar: Nice! Again, I think you could use `std::ranges::accumulate` instead, but it doesn't matter much.

src/HealthGPS/analysis_module.cpp
```diff
@@ -26,44 +26,38 @@ SimulationModuleType AnalysisModule::type() const noexcept {
...
+
+    // Set the vector size and initialise all values to 0.0
+    calculated_factors_.resize(total_num_bins * num_factors_to_calc, 0.0);
```
alexdewar, line 61: Zero is the default value anyway, so it's fine to just do this:

```suggestion
    calculated_factors_.resize(total_num_bins * num_factors_to_calc);
```

alexdewar (approved):
A few comments, but LGTM!

src/HealthGPS/analysis_module.cpp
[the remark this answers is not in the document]
TinyMarsh: It is now!

TinyMarsh:
Okay I think this is good to go now, any final thoughts @alexdewar @jamesturner246 ?

There's a lingering clang-tidy suggestions, which I thought I fixed by using a transparent functor, but it's still not happy. Perhaps I'm just not understanding what it's complaining about.

jamesturner246 (approved):
LGTM :+1: 

Might just be an outdated check from last commit :shrug:

PR #439, opened 2024-06-11
StdDev for YLL, YLD, DALY.

### LINKED ISSUE #438: Compute standard deviation of more variables
Following up #326, compute standard deviation of YLL, YLD and DALY.

jamesturner246:
Fixes #438. 

Following up from the StdDev changes, we now compute them for YLL, YLD and DALY. 

Original fields are now named `mean_ylll`, etcetera. Update your scripts accordingly. @jzhu20 @jackolney, please have a quick glance through the code to make sure you understand, and please do a little test run. I think it's important that you know what these values represent.

On that note, I really think this YLL calculation is wrong, by the way. It takes the sum of all theoretical years lost from recently deal people, but *divides it by the total number of active people*. Surely it should be divided by the number of dead people at that year. We should discuss this soon, as now is the perfect time to fix this.


src/HealthGPS/analysis_module.cpp
```diff
@@ -385,43 +388,64 @@ void AnalysisModule::calculate_population_statistics(RuntimeContext &context,
...
 // NOLINTEND(readability-function-cognitive-complexity)
 
-void AnalysisModule::calculate_standard_deviation(RuntimeContext &context, DataSeries &series) {
-    const int min_age = context.age_range().lower();
-    const int max_age = context.age_range().upper();
+void AnalysisModule::calculate_standard_deviation(RuntimeContext &context,
+                                                  DataSeries &series) const {
 
+    // Accumulate squared deviations from mean.
+    auto std_accumulate = [&series](const std::string &chan, core::Gender sex, int age,
```
alexdewar, line 395: How about a slightly more descriptive name, like `accumulate_square_diffs` or something?

src/HealthGPS/analysis_module.cpp
```diff
@@ -385,43 +388,64 @@ void AnalysisModule::calculate_population_statistics(RuntimeContext &context,
...
-            const double mean = series(sex, "mean_" + factor.key().to_string()).at(age);
-            const double diff = value - mean;
-            series(sex, "std_" + factor.key().to_string()).at(age) += diff * diff;
+            std_accumulate(factor.key().to_string(), sex, age, value);
         }
     }
 
     // Calculate in-place standard deviation.
-    for (int i = min_age; i <= max_age; i++) {
+    auto std_compute = [&series](const std::string &chan, core::Gender sex, int age) {
```
alexdewar, line 430: Ditto above. `divide_by_count` or something?

alexdewar (approved):
A couple of comments re naming, but otherwise all good 😄

src/HealthGPS/analysis_module.cpp
jamesturner246: Done :+1:

src/HealthGPS/analysis_module.cpp
jamesturner246: Done :+1:

PR #444, opened 2024-06-17
Fix automerge workflow: use PR author not GitHub actor

### LINKED ISSUE #443: [BUG] Auto-merging of bot PRs broken
I borrowed @AdrianDAlessandro's config for automerging PRs from dependabot and the pre-commit bot. Bizarrely, the workflow only also seems to be activated for other PRs though. See e.g. my PR here: https://github.com/imperialCHEPI/healthgps/actions/runs/9498223239/job/26176474545?pr=440

The config file is here and I think it should be identical to the one in the `pip-tools-template` repo: https://github.com/imperialCHEPI/healthgps/blob/main/.github/workflows/auto-merge.yml

Any ideas @AdrianDAlessandro?

I've disabled auto-merges for now while we figure this out.

alexdewar:
Currently auto-merge is being automatically enabled for some non-bot PRs.

@AdrianDAlessandro has pointed out that this is because the pre-commit workflow is automatically pushing fixes to our PRs and, in that case, the last committer (the "GitHub actor") will be the pre-commit bot, so this workflow is being erroneously activated.

It seems like the right way to fix this is to use the PR's author instead: https://github.com/orgs/community/discussions/25502

Fixes #443.

### DIFF
--- a/.github/workflows/auto-merge.yml
+++ b/.github/workflows/auto-merge.yml
@@ -10,7 +10,7 @@ permissions:
 jobs:
   auto-merge:
     runs-on: ubuntu-latest
-    if: ${{ github.actor == 'dependabot[bot]' || github.actor == 'pre-commit-ci[bot]' }}
+    if: ${{ github.event.pull_request.user.login == 'dependabot[bot]' || github.event.pull_request.user.login == 'pre-commit-ci[bot]' }}
     steps:
       - name: Enable auto-merge for Dependabot PRs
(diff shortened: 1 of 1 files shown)


jamesturner246 (approved):
Okay, so `github.actor` is the last committer, which might be pre-commit, even in one of our PRs, so we can't use actor to trigger automerge. Makes sense. LGTM :+1:

alexdewar:
@jamesturner246 Yep, exactly.

Issue #407, opened 2024-05-31
Data store source should be specified in model config file


alexdewar:
As the version of the data files used is a crucial part of the model run, I think really the path to the data source should be in the config file rather than specified as a command-line argument (or better still, as a URL: #364).

Do you agree, @jamesturner246?

jamesturner246:
Sounds sensible! :+1:

PR #445, opened 2024-06-17
Allow for downloading zipped input data from URL

### LINKED ISSUE #364: Allow for reading data from URL

alexdewar:
This PR adds support for entering a URL to the input data instead of a file path on the command line. I implemented this using a C++ wrapper for curl.

Note that files are downloaded to a temporary folder and that there is no caching of URLs, just caching to check whether a zip file matching a particular checksum has already been extracted. While we could implement this by allowing users to pass a checksum for the input data as a command line argument, I don't think that's particularly user friendly. The better option is to let users specify it via the config file (#407), which I'll do later.

Closes #364.

### DIFF
--- a/src/HealthGPS.Console/command_options.cpp
+++ b/src/HealthGPS.Console/command_options.cpp
@@ -61,17 +61,7 @@ CommandOptions parse_arguments(cxxopts::Options &options, int &argc, char *argv[
         }
 
         if (result.count("storage")) {
-            cmd.storage_folder = result["storage"].as<std::string>();
-            if (cmd.storage_folder.is_relative()) {
-                cmd.storage_folder = std::filesystem::absolute(cmd.storage_folder);
-                fmt::print("File storage folder.: {}\n", cmd.storage_folder.string());
-            }
-        }
-
-        if (!fs::exists(cmd.storage_folder)) {
-            fmt::print(fg(fmt::color::red), "\nFile storage folder: {} not found.\n",
-                       cmd.storage_folder.string());
-            cmd.exit_code = EXIT_FAILURE;
+            cmd.data_path_or_url = result["storage"].as<std::string>();
         }
 
         if (result.count("jobid")) {
--- a/src/HealthGPS.Console/command_options.cpp
+++ b/src/HealthGPS.Console/command_options.cpp
@@ -62,6 +62,17 @@ CommandOptions parse_arguments(cxxopts::Options &options, int &argc, char *argv[
 
         if (result.count("storage")) {
             cmd.data_path_or_url = result["storage"].as<std::string>();
+
+            if (!cmd.data_path_or_url.starts_with("http://") &&
+                !cmd.data_path_or_url.starts_with("https://")) {
+                const std::filesystem::path path = cmd.data_path_or_url;
+
+                if (path.is_relative()) {
+                    cmd.data_path_or_url = std::filesystem::absolute(path).string();
+                }
+            }
+
+            fmt::print("Data source: {}\n", cmd.data_path_or_url);
         }
 
         if (result.count("jobid")) {
--- a/src/HealthGPS.Console/command_options.h
+++ b/src/HealthGPS.Console/command_options.h
@@ -20,7 +20,7 @@ struct CommandOptions {
     std::filesystem::path config_file{};
 
     /// @brief The back-end storage full path or URL argument value
-    std::string data_path_or_url{};
+    std::string data_path_or_url;
 
     /// @brief Indicates whether the application logging is verbose
     bool verbose{};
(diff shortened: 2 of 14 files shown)


src/HealthGPS.Console/command_options.h
```diff
@@ -19,8 +19,8 @@ struct CommandOptions {
     /// @brief The configuration file argument value
     std::filesystem::path config_file{};
 
-    /// @brief The back-end storage full path argument value
-    std::filesystem::path storage_folder{};
+    /// @brief The back-end storage full path or URL argument value
+    std::string data_path_or_url;
```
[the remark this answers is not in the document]
alexdewar: Isn't that what I just wrote? :thinking:

src/HealthGPS.Console/command_options.cpp
```diff
@@ -61,17 +61,7 @@ CommandOptions parse_arguments(cxxopts::Options &options, int &argc, char *argv[
...
-                cmd.storage_folder = std::filesystem::absolute(cmd.storage_folder);
-                fmt::print("File storage folder.: {}\n", cmd.storage_folder.string());
-            }
-        }
-
-        if (!fs::exists(cmd.storage_folder)) {
-            fmt::print(fg(fmt::color::red), "\nFile storage folder: {} not found.\n",
-                       cmd.storage_folder.string());
-            cmd.exit_code = EXIT_FAILURE;
+            cmd.data_path_or_url = result["storage"].as<std::string>();
```
jamesturner246, line 64: I guess it doesn't make sense any more to use a relative path in the config file?

src/HealthGPS.Datastore/datamanager.cpp
```diff
@@ -46,15 +47,22 @@ nlohmann::json read_input_files_from_directory(const std::filesystem::path &root
...
-            fmt::format("Path must either point to a zip file or a directory: {}", path.string()));
+        std::filesystem::path path = path_or_url;
+        if (std::filesystem::is_directory(path)) {
+            root_ = std::move(path);
+        } else if (std::filesystem::is_regular_file(path) && path.extension() == ".zip") {
+            root_ = extract_zip_file_or_load_from_cache(path);
+        } else {
+            throw std::runtime_error(fmt::format(
+                "Path must either point to a zip file or a directory: {}", path_or_url));
+        }
```
jamesturner246, line 65: Will this implicitly check if the file exists?

src/HealthGPS.Datastore/download_file.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+namespace hgps::data {
+std::filesystem::path download_file(const std::string &url,
+                                    const std::filesystem::path &download_path) {
+    std::ofstream ofs{download_path};
+    if (!ofs) {
+        throw std::runtime_error(fmt::format("Failed to create file {}", download_path.string()));
+    }
+
+    curlpp::Cleanup cleanup;
```
jamesturner246, line 42: What does this do?

jamesturner246 (approved):
LGTM :+1:  Only questions.

src/HealthGPS.Console/command_options.cpp
alexdewar: No, it does! I must have just deleted this without looking too closely at it. Oops :see_no_evil: 

I'll fix it up.

src/HealthGPS.Datastore/download_file.cpp
alexdewar: Cleans up curl's global resources when it goes out of scope. Uses this function: https://curl.se/libcurl/c/curl_global_cleanup.html

alexdewar: It looks a bit weird but I guess that's the C++/RAII way!

src/HealthGPS.Console/command_options.cpp
alexdewar: Oh wait. This is a bit weird. It converts relative paths to absolute ones, but it doesn't actually change what the path points to. I assumed it was treating relative paths as if they were relative to the config file's path (and then fixing them up), but they're actually just relative to the CWD.

I suppose either way, if users are expecting to see absolute paths in log files etc., we should probably keep this behaviour though. What do you think?

If we were writing this from scratch, I think a better behaviour would be for paths to be relative to the config file, but I guess we'll break things if we change it now.

jamesturner246: > I suppose either way, if users are expecting to see absolute paths in log files etc., we should probably keep this behaviour though. What do you think?

Yeah, probably for the best. Rather than fiddling around with data dir location and config relative paths -- sounds error prone.

alexdewar: It's easy enough to do, but I don't want to break people's existing config files.
Hopefully going forward everyone will use URLs instead anyway!

PR #448, opened 2024-06-18
Fix: Bot PRs need approving review before they can be merged

alexdewar:
I missed this before, but because we have it as a requirement that PRs need an approving review, the bot PRs are still not being auto-merged.

This change should fix that.


jamesturner246 (approved):
Excellent! Now if only there was a way to filter the tens of emails for bumping various versions each week :thinking:

Issue #420, opened 2024-06-06
analysis code to compute group mean for arbitrary variables


jamesturner246:
Depends on #419.

Add code to analysis module for looping over arbitrary variables (just do two for now), and computing their mean.

Hopefully this will be as simple as replacing `gender` and `age` with the new variables.

Just have constant variable keys for now -- we can worry about feeding it in via config later.

PR #453, opened 2024-06-19
Implement new `calculate_population_statistics` method.

TinyMarsh:
This begins to tackle #420, but does not completely close that issue because I wanted to keep the PRs at at a reasonably-digestible size. More PRs to come if/when this gets approved. Happy to discuss this PR in person as it's a bit weird.

- Move initialisation of `factors` to the header file as a data member and rename to `factors_to_calculate_`. Eventually this will be set elsewhere anyway.
- Move `factor_bins_` to the header file as a data member. This is because it needs to be accessed elsewhere in the class.
- Introduces `factor_bin_widths_` as a data member. This helps us calculate which bin a given factor value is in later on.
- Overload the `calculate_population_statistics` function with the `calculated_factors_` vector in the signature instead of the `series` object. We can remove the original function when we have completed implementation of this new way of calculating the analysis.

The new, overloaded `calculate_population_statistics` function is the main focus of this PR. At the moment all it is doing is calculating the correct index in the `calculated_factors_` vector for a given entity. This is necessary because we are representing a matrix of arbitrary dimensions in a flat vector. The number of dimensions is equal to the size of `factors_to_calculate_`.

The calculation for fetching the correct index is based on the following idea for calculating the index in a 3D, row-major ordered matrix with dimensions of length `L`, `M`, and `N`.
```
matrix[ i ][ j ][ k ] = vector[ i*(N*M) + j*M + k ]
```
but extended to be able to deal with an arbitrary number of dimensions (I *think* I have done this right, really this is the bit of the PR that I'm most interested in being reviewed).

### DIFF
--- a/src/HealthGPS/analysis_module.cpp
+++ b/src/HealthGPS/analysis_module.cpp
@@ -24,12 +24,11 @@ AnalysisModule::AnalysisModule(AnalysisDefinition &&definition, WeightModel &&cl
 
 // Overload constructor with additional parameter for calculated_factors_
 AnalysisModule::AnalysisModule(AnalysisDefinition &&definition, WeightModel &&classifier,
-                               const core::IntegerInterval age_range, unsigned int comorbidities, std::vector<double> calculated_factors)
+                               const core::IntegerInterval age_range, unsigned int comorbidities,
+                               std::vector<double> calculated_factors)
     : definition_{std::move(definition)}, weight_classifier_{std::move(classifier)},
       residual_disability_weight_{create_age_gender_table<double>(age_range)},
-      comorbidities_{comorbidities},
-        calculated_factors_{std::move(calculated_factors)}
-{}
+      comorbidities_{comorbidities}, calculated_factors_{std::move(calculated_factors)} {}
 SimulationModuleType AnalysisModule::type() const noexcept {
     return SimulationModuleType::Analysis;
 }
@@ -338,14 +337,12 @@ void AnalysisModule::calculate_population_statistics(RuntimeContext &context) co
         // Calculate the index in the calculated_factors_ vector
         auto index = 0;
         for (size_t i = 0; i < bin_indices.size() - 1; i++) {
-            auto accumulated_bins = std::accumulate(std::next(factor_bins_.cbegin(), i + 1), factor_bins_.cend(), 1, std::multiplies<>());
+            auto accumulated_bins = std::accumulate(std::next(factor_bins_.cbegin(), i + 1),
+                                                    factor_bins_.cend(), 1, std::multiplies<>());
             index +=
-                static_cast<size_t>(bin_indices[i] *
-                accumulated_bins *
-                num_factors_to_calculate);
+                static_cast<size_t>(bin_indices[i] * accumulated_bins * num_factors_to_calculate);
         }
-        index +=
-            static_cast<size_t>(bin_indices.back() * num_factors_to_calculate);
+        index += static_cast<size_t>(bin_indices.back() * num_factors_to_calculate);
 
         // Now we can add the values of the factors that are not in factors_to_calculate_
         for (const auto &factor : context.mapping().entries()) {
--- a/src/HealthGPS/analysis_module.cpp
+++ b/src/HealthGPS/analysis_module.cpp
@@ -319,15 +319,10 @@ void AnalysisModule::calculate_population_statistics(RuntimeContext &context) {
         context.mapping().entries().size() - factors_to_calculate_.size();
 
     for (const auto &person : context.population()) {
-        // Get factors_to_calculate_ values for this entity
-        std::vector<double> person_factors;
-        for (const auto &factor : factors_to_calculate_) {
-            person_factors.push_back(person.get_risk_factor_value(factor));
-        }
         // Get the bin index for each factor
         std::vector<size_t> bin_indices;
         for (size_t i = 0; i < factors_to_calculate_.size(); i++) {
-            auto factor_value = person_factors[i];
+            double factor_value = person.get_risk_factor_value(factors_to_calculate_[i]);
             size_t bin_index =
                 static_cast<size_t>((factor_value - factor_min_values_[i]) / factor_bin_widths_[i]);
             bin_indices.push_back(bin_index);
--- a/src/HealthGPS/analysis_module.cpp
+++ b/src/HealthGPS/analysis_module.cpp
@@ -66,7 +66,7 @@ void AnalysisModule::initialise_vector(RuntimeContext &context) {
     // The product of the number of bins for each factor can be used to calculate the size of the
     // `calculated_factors_` in the next step
     size_t total_num_bins =
-        std::accumulate(factor_bins_.cbegin(), factor_bins_.cend(), 1, std::multiplies<>());
+        std::accumulate(factor_bins_.cbegin(), factor_bins_.cend(), 1u, std::multiplies<>());
 
     // Set the vector size and initialise all values to 0.0
     calculated_factors_.resize(total_num_bins * num_factors_to_calc);
--- a/src/HealthGPS/analysis_module.cpp
+++ b/src/HealthGPS/analysis_module.cpp
@@ -37,6 +37,7 @@ SimulationModuleType AnalysisModule::type() const noexcept {
 void AnalysisModule::initialise_vector(RuntimeContext &context) {
     factor_bins_.reserve(factors_to_calculate_.size());
     factor_bin_widths_.reserve(factors_to_calculate_.size());
+    factor_min_values_.reserve(factors_to_calculate_.size());
 
     for (const auto &factor : factors_to_calculate_) {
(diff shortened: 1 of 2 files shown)


src/HealthGPS/analysis_module.cpp
```diff
@@ -304,6 +305,47 @@ DALYsIndicator AnalysisModule::calculate_dalys(Population &population, unsigned
...
+void AnalysisModule::calculate_population_statistics(RuntimeContext &context) const {
+    auto num_factors_to_calculate =
+        context.mapping().entries().size() - factors_to_calculate_.size();
+
+    for (const auto &entity : context.population()) {
+        // Get factors_to_calculate_ values for this entity
+        std::vector<double> entity_factors;
+        for (const auto &factor : factors_to_calculate_) {
+            entity_factors.push_back(entity.get_risk_factor_value(factor));
+        }
```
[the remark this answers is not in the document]
alexdewar: I'd actually define `entity_factors` outside the loop above this one, then `clear()` it on every iteration. That way you don't need to create a new vector every time. It's a bit of a micro-optimisation though

src/HealthGPS/analysis_module.cpp
```diff
@@ -304,6 +305,47 @@ DALYsIndicator AnalysisModule::calculate_dalys(Population &population, unsigned
                           .disability_adjusted_life_years = yll + yld};
 }
 
+void AnalysisModule::calculate_population_statistics(RuntimeContext &context) const {
+    auto num_factors_to_calculate =
+        context.mapping().entries().size() - factors_to_calculate_.size();
+
+    for (const auto &entity : context.population()) {
```
alexdewar, line 312: Maybe call this variable `person` rather than `entity`? It's not your fault -- it looks like there are loads of `entity`s in this file (and throughout the code base) -- but it did take me a while to twig that it referred to a person

src/HealthGPS/analysis_module.cpp
```diff
@@ -304,6 +305,47 @@ DALYsIndicator AnalysisModule::calculate_dalys(Population &population, unsigned
...
+        std::vector<double> entity_factors;
+        for (const auto &factor : factors_to_calculate_) {
+            entity_factors.push_back(entity.get_risk_factor_value(factor));
+        }
+        // Get the bin index for each factor
+        std::vector<int> bin_indices;
+        for (size_t i = 0; i < factors_to_calculate_.size(); i++) {
+            auto factor_value = entity_factors[i];
+            auto bin_index =
+                static_cast<int>((factor_value - factor_bin_widths_[i]) / factor_bin_widths_[i]);
```
alexdewar, line 323: I'm sure I'm missing something here, but why are you subtracting `factor_bin_widths_[i]`? Shouldn't we be subtracting the minimum possible factor value? Or do they all start at 0?

alexdewar (changes requested):
A few small suggestions and a question, but otherwise looks good.

I tried reasoning through the code for calculating the indices but it made my head hurt after a while :laughing:. It didn't look obviously wrong but I'm not confident that I've got it right either! How about making a separate standalone function for caculating indices for an n-dimensional matrix and writing a unit test for it (using some hand-calculated values)?

src/HealthGPS/analysis_module.cpp
```diff
@@ -304,6 +305,47 @@ DALYsIndicator AnalysisModule::calculate_dalys(Population &population, unsigned
...
+    for (const auto &entity : context.population()) {
+        // Get factors_to_calculate_ values for this entity
+        std::vector<double> entity_factors;
+        for (const auto &factor : factors_to_calculate_) {
+            entity_factors.push_back(entity.get_risk_factor_value(factor));
+        }
+        // Get the bin index for each factor
+        std::vector<int> bin_indices;
+        for (size_t i = 0; i < factors_to_calculate_.size(); i++) {
+            auto factor_value = entity_factors[i];
```
jamesturner246, line 321: I think this vector can be removed entirely:
```suggestion
        // Get the bin index for each factor
        std::vector<int> bin_indices;
        for (size_t i = 0; i < factors_to_calculate_.size(); i++) {
            double factor_value = entity.get_risk_factor_value(factors_to_calculate_[i]);
```

src/HealthGPS/analysis_module.cpp
```diff
@@ -304,6 +305,47 @@ DALYsIndicator AnalysisModule::calculate_dalys(Population &population, unsigned
...
+        }
+
+        // Calculate the index in the calculated_factors_ vector
+        auto index = 0;
+        for (size_t i = 0; i < bin_indices.size() - 1; i++) {
+            index +=
+                static_cast<size_t>(bin_indices[i]) *
+                static_cast<size_t>(std::accumulate(std::next(factor_bins_.cbegin(), i + 1),
+                                                    factor_bins_.cend(), 1, std::multiplies<>())) *
+                static_cast<size_t>(num_factors_to_calculate);
```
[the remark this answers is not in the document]
jamesturner246: Yup. Can also just have all size related integers as `size_t` in the init and header.

src/HealthGPS/analysis_module.cpp
jamesturner246: I'm a bit puzzled by this also. Isn't `factor_bin_widths` a real number `r >= 1.0`? Surely `factor_value` would also need normalising before use?

jamesturner246 (changes requested):
Starting to take shape :+1:. Some fixing up to do still.

src/HealthGPS/analysis_module.cpp
jamesturner246: Eventually, given the fixed window of bins, we will need to make sure all our risk factors have config supplied boundaries, which are always respected. Then we can just read `min_value` and `max_value` directly from that and always be safe. But another time...

TinyMarsh: Good point and well spotted. I have no idea why I had it like originally. Likely because at some point I changed the logic and missed it.

src/HealthGPS/analysis_module.h
```diff
@@ -23,6 +23,16 @@ class AnalysisModule final : public UpdatableModule {
...
                    const core::IntegerInterval age_range, unsigned int comorbidities);
 
+    /// @brief Initialises a new instance of the AnalysisModule class.
+    /// @param definition The analysis module definition
+    /// @param classifier Body weight classifier model instance
+    /// @param age_range The experiment valid age range
+    /// @param comorbidities Maximum number of comorbidities to include
+    /// @param calculated_factors The calculated factors to include
+    AnalysisModule(AnalysisDefinition &&definition, WeightModel &&classifier,
+                   const core::IntegerInterval age_range, unsigned int comorbidities,
```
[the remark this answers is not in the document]
TinyMarsh: Meh I kind of disagree with this rule, but whatever.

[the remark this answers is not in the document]
TinyMarsh: Meh I kind of disagree with this rule, but whatever.

src/HealthGPS/analysis_module.cpp
```diff
@@ -304,6 +314,39 @@ DALYsIndicator AnalysisModule::calculate_dalys(Population &population, unsigned
...
+void AnalysisModule::calculate_population_statistics(RuntimeContext &context) {
+    size_t num_factors_to_calculate =
+        context.mapping().entries().size() - factors_to_calculate_.size();
+
+    for (const auto &person : context.population()) {
+        // Get the bin index for each factor
+        std::vector<size_t> bin_indices;
+        for (size_t i = 0; i < factors_to_calculate_.size(); i++) {
+            double factor_value = person.get_risk_factor_value(factors_to_calculate_[i]);
+            size_t bin_index =
```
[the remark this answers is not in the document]
TinyMarsh: Why though?

TinyMarsh:
Thanks for the feedback @jamesturner246 @alexdewar . This should be good for another review now.

I don't really understand the wall of text output for why the build is failing on Windows. I think it doesn't like this:
https://github.com/imperialCHEPI/healthgps/blob/b596d0b010073749e39ccd540a35c85869aca390/src/HealthGPS/analysis_module.cpp#L68-L69
but I'm not really sure what it wants me to do. Any ideas?

jamesturner246:
Humm.. Try:
```c++
 size_t total_num_bins = 
     std::accumulate(factor_bins_.cbegin(), factor_bins_.cend(), 1u, std::multiplies<>()); 
```
Note `1u` for unsigned int literal.

jamesturner246 (approved):
This looks good. :+1: 

Re. the Windows bug, looks like a pedantic casting warning, Try using `1u` as mentioned.

alexdewar (approved):
LGTM (except for that weird MSVC failure).

TinyMarsh:
Yep well done @jamesturner246 that was the problem, I ended up having to make the accumulator type in `std::accumulate` match the type of `total_num_bins`. MSVC seems happy now.

PR #455, opened 2024-06-20
Fix YLL/Y:D/DALY and death

### LINKED ISSUE #442: Fix incorrect YLL and deaths in output
Right now, YLL is computed as `sum_over_dead_pop(years_lost_for_person * 100000) / n_alive`, but this doesn't seem like 'true' YLL. Surely it's divided by either `n_dead` or `n_dead + n_alive` or something.

If Health-GPS states that the output value is YLL, then it should be YLL, not some intermediate value.

We need to figure this out and output something sensible.

jamesturner246:
Fixes #442. 

The loops for mean and std have been tidied up, and are now 'opt-in', so no unwanted surprises in other columns that don't opt in.

YLL/YLD/DALY are now correctly computed by dividing by active *PLUS* deceased population.

Everything that should be an integer, such as deaths, migrations, weight classifications, is now an integer. Update your scripts accordingly.

### DIFF
--- a/src/HealthGPS/analysis_module.cpp
+++ b/src/HealthGPS/analysis_module.cpp
@@ -307,8 +307,6 @@ DALYsIndicator AnalysisModule::calculate_dalys(Population &population, unsigned 
 // NOLINTBEGIN(readability-function-cognitive-complexity)
 void AnalysisModule::calculate_population_statistics(RuntimeContext &context,
                                                      DataSeries &series) const {
-    auto min_age = context.age_range().lower();
-    auto max_age = context.age_range().upper();
     if (series.size() > 0) {
         throw std::logic_error("This should be a new object!");
     }
@@ -361,7 +359,8 @@ void AnalysisModule::calculate_population_statistics(RuntimeContext &context,
     }
 
     // For each age group in the analysis...
-    for (int age = min_age; age <= max_age; age++) {
+    const auto age_range = context.age_range();
+    for (int age = age_range.lower(); age <= age_range.upper(); age++) {
         double count_F = series(core::Gender::female, "count").at(age);
         double count_M = series(core::Gender::male, "count").at(age);
         double deaths_F = series(core::Gender::female, "deaths").at(age);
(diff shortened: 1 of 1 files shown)


jamesturner246:
Don't they do CI on these Ubuntu builder images? :confused:

jzhu20:
I don't really understand the codes for calculating standard deviation.
Lines 452-456 and 458-461: Do you substract individual value from group mean, then take squared to generate sum? I cannot locate this step. I get sqrt(sum/count) or sqrt(sum/(count+deaths)).

jamesturner246:
Hi @jzhu20, this was done in a prior PR -- see lines 402-434.

jzhu20 (approved):

jamesturner246:
Any thoughts @alexdewar?

alexdewar:
Sorry for the slow reply! I'll take a look at this today

src/HealthGPS/analysis_module.cpp
```diff
@@ -307,29 +307,28 @@ DALYsIndicator AnalysisModule::calculate_dalys(Population &population, unsigned
...
-        auto age = entity.age;
-        auto gender = entity.gender;
+    for (const auto &person : context.population()) {
+        auto age = person.age;
+        auto gender = person.gender;
 
-        if (!entity.is_active()) {
-            if (!entity.is_alive() && entity.time_of_death() == current_time) {
+        if (!person.is_active()) {
+            if (!person.is_alive() && person.time_of_death() == current_time) {
```
alexdewar, line 322: Not relevant for this PR, but I feel like a `has_died()` method would be clearer than an `is_alive()` one.

src/HealthGPS/analysis_module.cpp
```diff
@@ -487,9 +502,6 @@ void AnalysisModule::initialise_output_channels(RuntimeContext &context) {
         channels_.emplace_back("incidence_" + disease.code.to_string());
     }
 
-    channels_.emplace_back("disability_weight");
```
alexdewar, line 490: Don't we need `disability_weight` anymore?

src/HealthGPS/analysis_module.cpp
```diff
@@ -307,29 +307,28 @@ DALYsIndicator AnalysisModule::calculate_dalys(Population &population, unsigned
 // NOLINTBEGIN(readability-function-cognitive-complexity)
 void AnalysisModule::calculate_population_statistics(RuntimeContext &context,
                                                      DataSeries &series) const {
-    auto min_age = context.age_range().lower();
-    auto max_age = context.age_range().upper();
     if (series.size() > 0) {
         throw std::logic_error("This should be a new object!");
     }
```
alexdewar, line 312: I think it would make more sense to return a new `DataSeries` object rather than modify an input argument.

But that's not something we need to do now.

alexdewar (approved):
I've eyeballed the code and it looks like it's doing the right thing. I've made some comments, none of which are actually about this PR :upside_down_face: 

That said, the `calculate_population_statistics()` method is a bit overcomplicated (we have to suppress a `clang-tidy` warning about this), so trying to keep track of everything that's going on made my head hurt. We should probably factor out some of the logic into separate functions at some point. Maybe we need a new issue for this?

jamesturner246:
> That said, the `calculate_population_statistics()` method is a bit overcomplicated (we have to suppress a `clang-tidy` warning about this), so trying to keep track of everything that's going on made my head hurt. We should probably factor out some of the logic into separate functions at some point. Maybe we need a new issue for this?

Agree in principle, but I'm hesitant to invest any more time in this whilst knowing it will eventually be replaced by @TinyMarsh's work. Let's wait for the result of that before making any decisions.

src/HealthGPS/analysis_module.cpp
jamesturner246: Then you'd have `is_alive` and `has_died =  !is_alive` functions?

Unless you mean like:

```c++
bool Person::has_just_died(unsigned int current_time) {
    return !is_alive() && time_of_death() == current_time;
}
```

PR #456, opened 2024-06-24
Fix race conditions when extracting zip files

### LINKED ISSUE #454: Make zip caching code robust to race conditions
Currently, we extract zip files to a cache folder, whose path is based on a hash of the zip file. The problem is that if another instance of HealthGPS is started while extraction is taking place, it may attempt to read the data while extraction is partway through.

This might sound unlikely, but it could definitely happen with multiple jobs running on the HPC simultaneously, as the same cache folder will be being used for each.

I think the right solution is to extract the zip file to a temporary folder (in the same parent folder as the destination folder) and then rename it once extraction is complete, which I think should be an atomic operation.

alexdewar:
Before extracting a zip file, we check whether we already have the contents of that file extracted to a cache folder and, if so, skip the extraction.

While this is safe to do if there is only one HealthGPS process running, if there are multiple processes attempting to extract the zip file at the same time, there is a potential race condition (#454):

1. Process A observes that the cache folder doesn't exist, creates it and begins extraction
2. Process B observes that the cache folder exists, tries to use it before extraction is complete and crashes
3. Process A completes extraction

Note that while this sounds unlikely, it is entirely possible that this could happen with multiple jobs running on the HPC (particularly given that filesystem access is often slow on the Imperial HPC).

Another possible bug is that a HealthGPS could terminate before extraction is complete, in which case, again, the cache folder will exist, but only contain a subset of the data files. In this case, the only remedy would be to delete the cache folder -- not particularly user friendly.

This PR solves both of these problems by performing extraction to a temporary folder (with a unique name) before renaming the folder. While moving a folder is not guaranteed to be atomic because it may involve copying data between devices, it seems that renaming them *should* be (or at least it is on POSIX, which will fix the HPC case).

Fixes #454.

### DIFF
--- a/src/HealthGPS.Datastore/zip_file.cpp
+++ b/src/HealthGPS.Datastore/zip_file.cpp
@@ -6,7 +6,28 @@
 #include <libzippp.h>
 
 #include <fstream>
-#include <random>
+
+namespace {
+/// @brief Create a temporary directory for extracting a zip file into
+/// @param output_path The final output path where files will be moved to
+/// @return The path to the temporary directory
+std::filesystem::path create_temporary_extract_directory(const std::filesystem::path &output_path) {
+    // TODO: Randomise temp path
+    std::filesystem::path temp_path = output_path.string() + ".tmp";
+
+    // Perhaps extraction was stopped halfway through? Delete it and start over
+    if (std::filesystem::exists(temp_path)) {
+        std::filesystem::remove_all(temp_path);
+    }
+
+    if (!std::filesystem::create_directories(temp_path)) {
+        throw std::runtime_error(
+            fmt::format("Failed to create temporary directory: {}", temp_path.string()));
+    }
+
+    return temp_path;
+}
+} // anonymous namespace
 
 namespace hgps::data {
 std::filesystem::path get_zip_cache_directory(const std::string &file_hash) {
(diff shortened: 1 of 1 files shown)


src/HealthGPS.Datastore/zip_file.cpp
```diff
@@ -52,11 +100,6 @@ std::filesystem::path extract_zip_file_or_load_from_cache(const std::filesystem:
         return cache_path;
     }
```
jamesturner246, line 101:
```c++
    if (!std::filesystem::exists(cache_path)) {
        extract_zip_file(file_path, cache_path);
    }
    return cache_path;
```

jamesturner246 (approved):
Looks good. just a nitpick.

TinyMarsh (approved):
Nice! LGTM.

I don't know if this is relevant but would [std::filesystem::temp_directory_path](https://en.cppreference.com/w/cpp/filesystem/temp_directory_path) be useful here? Perhaps there's a reason we wouldn't want to use the system temp directory?

alexdewar:
@TinyMarsh Ordinarily that would make sense, but the weird thing about what we're doing here is that we want to be able to move the temporary folder to its final destination atomically (i.e. so that another HPC job doesn't try to read the cache folder when it's only got half the files in) and moving a folder will only be atomic if it's in the same partition, which is why we do it in the same cache folder (e.g. on Linux `/tmp` is usually a ramdisk, which means the files would have to be copied).

PR #458, opened 2024-06-24
Allow HTTP redirects for file downloads

### LINKED ISSUE #457: Passing URL of data source from GitHub doesn't work
## Describe the bug

I'm currently trying to use a GitHub URL as a data source without success. Specifically, I'm using the URL to a release asset for the `healthgps-data` repo: https://github.com/imperialCHEPI/healthgps-data/releases/download/20240624/data_20240624.zip

We just get an empty file in response, which obviously breaks things.

The zip file itself actually does have files in it though!

alexdewar:
It turns out GitHub often redirects file downloads to another URL, which the current code won't handle. Fix by telling libcurl to allow this (it's disabled by default).

Fixes #457.

With this fix, we can finally use data from the `healthgps-data` repo as an input source directly :smile:

See: https://github.com/imperialCHEPI/healthgps-data/releases/tag/20240624

### DIFF
--- a/src/HealthGPS.Datastore/download_file.cpp
+++ b/src/HealthGPS.Datastore/download_file.cpp
@@ -44,6 +44,7 @@ std::filesystem::path download_file(const std::string &url,
     // Our request to be sent
     curlpp::Easy request;
     request.setOpt<curlpp::options::Url>(url);
+    request.setOpt<curlpp::options::FollowLocation>(true);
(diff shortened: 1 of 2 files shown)


alexdewar:
PS -- I noticed download_file() was pointlessly returning an input argument, so I removed it.

jamesturner246 (approved):
LGTM :+1:

PR #463, opened 2024-06-26
Runner simplify

### LINKED ISSUE #462: Renaming of ModelRunner and HealthGPS classes
Rename some confusingly named things in HealthGPS, ModelRunner etc.

Remove needless Simulation interface.

jamesturner246:
Taming the beast 1/?. Straightforward change.

The only code changes are to remove the useless `Simulation` interface, rename `HealthGPS` to `Simulation` and to use start time from model definition instead of redundant context entry.

The remainder is renaming things.

Fixes #462.


src/HealthGPS/runner.cpp
```diff
@@ -107,27 +95,23 @@ double ModelRunner::run(Simulation &baseline, Simulation &intervention,
...
+void Runner::cancel() noexcept {
     if (is_running()) {
         source_.request_stop();
     }
 }
 
-void ModelRunner::run_model_thread(const std::stop_token &token, Simulation &model,
-                                   unsigned int run, const std::optional<unsigned int> seed) {
+void Runner::run_model_thread(const std::stop_token &token, Simulation &model, unsigned int run,
+                              const unsigned int seed) {
```
alexdewar, line 114: I have a vague feeling we changed `seed` to be an optional type at some point, because previously the code treated `seed==0` as meaning "don't seed the RNG" and using an optional seemed cleaner.

src/HealthGPS/simulation.cpp
```diff
@@ -113,27 +96,27 @@ adevs::Time HealthGPS::update(adevs::SimEnv<int> *env) {
...
 }
 
-adevs::Time HealthGPS::update(adevs::SimEnv<int> * /*env*/, std::vector<int> & /*x*/) {
+adevs::Time Simulation::update(adevs::SimEnv<int> * /*env*/, std::vector<int> & /*x*/) {
     // This method is never called because nobody sends messages.
     return adevs_inf<adevs::Time>();
 }
 
-void HealthGPS::fini(adevs::Time clock) {
+void Simulation::fini(adevs::Time clock) {
```
alexdewar, line 104: Is it just me or is `fini` a weird name for this method?

And if it's only ever called at the end of a simulation, maybe we should figure out how to do this using a destructor.

alexdewar (approved):
Except for my comment re using optional types for seeds (I might be wrong about this btw), LGTM!

A bit of spring cleaning is always good.

src/HealthGPS/simulation.cpp
jamesturner246: It's required/called by the strange ADEVS library. TBH we can just do away with it in future work. I can't see any benefit it introduces over making our own simple runner.

src/HealthGPS/runner.cpp
jamesturner246: I had that feeling also, but have checked the code and a seed is set for *all* simulations in `Runner::run` regardless. Maybe you were thinking of the RNG 'seed generator' RNG? The one in `program.cpp:137`, that generates seeds for job RNGs?

It's a bit confusing, hence the changes.

src/HealthGPS/simulation.cpp
alexdewar: Oh, yuck. Let's totally put that on the chopping board too

PR #464, opened 2024-06-28
Move config-related code into combined `HeathGPS.Input` module

### LINKED ISSUE #418: Move processing of all input files into same module
Currently the code for reading the static data lives in `HealthGPS.Datastore` while, for some reason, the code for reading model config files is in `HealthGPS.Console`. In order to implement #406, we will want to reuse some of the infrastructure in `HealthGPS.Datastore`, at which point it seems more logical to just have all of the input file parsing in the same place.

It would make more sense to have everything in one module, but `HealthGPS.Datastore` doesn't seem like the right name. @jamesturner246 Do you have any name preferences? We could go with `HealthGPS.InputFiles` or something.

alexdewar:
As per #418, all processing of input files can logically live in one module, so that's what I've done.

I've also changed all the code that's in the new `HealthGPS.Input` module to be inside a new `hgps::input` namespace. While I was at it I removed the strangely named `poco` namespace (replacing it with `hgps::input`) and just put the various bits in the `host` namespace into `hgps` (I was never sure what the point of the `host` namespace was and it seems redundant).

(There are a couple of trivial merge conflicts with #463, but I'm happy to fix those up after you've merged it @jamesturner246.)

Closes #418.


jamesturner246 (approved):
I love it all. Feel the weight just fall off. :+1:

PR #465, opened 2024-07-01
Allow users to specify data source in config file

### LINKED ISSUE #365: Check zipped input data matches a checksum
We should do this so that we can verify the integrity of downloaded data (#364).

### LINKED ISSUE #407: Data store source should be specified in model config file

alexdewar:
This PR adds support for passing in the path/URL to the data source via the config file, instead of as a command-line argument. For now, users can still use the command-line argument but it is deprecated and, if they do so, a warning is emitted. Longer term, I want to drop this option altogether as using the config file is a cleaner and more reproducible approach.

There are various types of data source the user can pass in:

1. A path to a data directory (as always)
2. A path to a zip file
3. A URL pointing to a zip file

If the user is supplying it via the config file, for the second two cases, we verify the data source against a checksum (not if it's a directory). For URLs, this means we can avoid downloading it altogether if the data is already in the cache. If they're doing it via the command-line argument, there's no checksum verification.

Given all these different permutations, the code is perhaps a little longer than I'd like, but I think it might be what's needed. We can drop some of it once we remove the option to supply the data source via the command line.

I haven't updated the docs yet, but will do so if people are happy with the changes.

Once this is merged, I plan to update the config files in the examples repo to use a URL as a data source.

Closes #407. Closes #365.

### DIFF
--- a/src/HealthGPS.Console/program.cpp
+++ b/src/HealthGPS.Console/program.cpp
@@ -112,8 +112,20 @@ int main(int argc, char *argv[]) { // NOLINT(bugprone-exception-escape)
 #ifdef CATCH_EXCEPTIONS
     try {
 #endif
+        // In future, we want users to supply the data source via the config file only, but for now
+        // we also allow passing it via a command line argument. Sanity check: Make sure they only
+        // do one of these things!
+        if (!(cmd_args.data_source.has_value() ^ config.data_source.has_value())) {
+            fmt::print(
+                fg(fmt::color::red),
+                "Must provide a data source via config file or command line, but not both\n");
+            return exit_application(EXIT_FAILURE);
+        }
+        const auto &data_source =
+            cmd_args.data_source.has_value() ? *cmd_args.data_source : *config.data_source;
+
         // Create back-end data store, cached data repository wrapper
-        auto data_api = input::DataManager(cmd_args.data_source, config.verbosity);
+        auto data_api = input::DataManager(data_source.get_data_directory(), config.verbosity);
         auto data_repository = hgps::CachedRepository{data_api};
 
         // Register the input risk factors model definitions
--- a/src/HealthGPS.Console/program.cpp
+++ b/src/HealthGPS.Console/program.cpp
@@ -113,7 +113,7 @@ int main(int argc, char *argv[]) { // NOLINT(bugprone-exception-escape)
     try {
 #endif
         // Create back-end data store, cached data repository wrapper
-        auto data_api = input::DataManager(cmd_args.data_path_or_url, config.verbosity);
+        auto data_api = input::DataManager(cmd_args.data_source, config.verbosity);
         auto data_repository = hgps::CachedRepository{data_api};
 
         // Register the input risk factors model definitions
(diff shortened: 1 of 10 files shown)


src/HealthGPS.Console/program.cpp
```diff
@@ -112,8 +112,23 @@ int main(int argc, char *argv[]) { // NOLINT(bugprone-exception-escape)
 #ifdef CATCH_EXCEPTIONS
     try {
 #endif
+        // In future, we want users to supply the data source via the config file only, but for now
+        // we also allow passing it via a command line argument. Sanity check: Make sure they only
+        // do one of these things!
+        if (!(cmd_args.data_source.has_value() ^ config.data_source.has_value())) {
```
TinyMarsh, line 118: I guess this could just be `if (A == B)`. But the XOR operator is cooler I suppose.

TinyMarsh (approved):
LGTM

src/HealthGPS.Console/program.cpp
alexdewar: Oh yeah :facepalm:. Good spot. I've done this somewhere else so I'll fix them both

src/HealthGPS.Input/data_source.cpp
```diff
@@ -0,0 +1,107 @@
...
+    const auto file_hash = hgps::compute_sha256_for_file(zip_file_path);
+
+    // If the cache folder already exists, then we don't need to extract anything
+    auto cache_path = hgps::input::get_zip_cache_directory(file_hash);
+    if (std::filesystem::is_directory(cache_path)) {
+        return cache_path;
+    }
+
+    // Extract files
+    hgps::input::extract_zip_file(zip_file_path, cache_path);
```
jamesturner246, line 73: I guess `extract_zip_file_or_load_from_cache` is no longer used?

src/HealthGPS.Input/data_source.cpp
```diff
@@ -0,0 +1,107 @@
...
+      validate_checksum_(true) {}
+
+std::filesystem::path DataSource::get_data_directory() const {
+    // If the data source is already a directory we can just return it (no checksum validation
+    // needed)
+    if (std::filesystem::is_directory(source_)) {
+        return source_;
+    }
+
+    if (!validate_checksum_) {
```
jamesturner246, line 97: Can the `validate_checksum_` bool be replaced with just `file_hash_.has_value()`?

jamesturner246 (changes requested):
Looks good, just a couple minor things.

src/HealthGPS.Input/data_source.cpp
alexdewar: True! I'll remove it

src/HealthGPS.Input/data_source.cpp
alexdewar: No :disappointed:. The problem is that we need to allow for the case where the user passes in the path to a zip file/URL via a command-line argument, in which case we don't require a checksum (we could, but it would be annoying for users).

jamesturner246: Ahh, I see. No worries.

PR #468, opened 2024-07-03
Implement stats calculations for new `calculated_stats_` container

TinyMarsh:
This PR adds much of the functionality to calculate the stats specified in `channels_`. A lot of the logic for this is copied from the existing code, but modified to use the `calculated_stats_` vector in place of the existing `series` object. As a reminder, this is necessary because the calculated stats are being stored in a flat vector (`calculated_stats_`) which represents a multidimensional array of factors that the user wants to group by, as opposed to the `series` object. A result of this is the introduction of the `calculate_index` method, and the `get_channel_index` method - both necessary to find the correct index of the factor bin, and "channel" stat in the `calculated_stats_`, respectively.

Note that this PR introduces a `accumulate_squared_diffs` method which is not being used yet. This will be used in the next PR tackling this body of work in which the standard deviation calculations will be completed.

### DIFF
--- a/src/HealthGPS/analysis_module.cpp
+++ b/src/HealthGPS/analysis_module.cpp
@@ -319,35 +319,87 @@ DALYsIndicator AnalysisModule::calculate_dalys(Population &population, unsigned 
 }
 
 void AnalysisModule::calculate_population_statistics(RuntimeContext &context) {
-    size_t num_factors_to_calculate =
-        context.mapping().entries().size() - factors_to_calculate_.size();
+
+    auto current_time = static_cast<unsigned int>(context.time_now());
 
     for (const auto &person : context.population()) {
-        // Get the bin index for each factor
-        std::vector<size_t> bin_indices;
-        for (size_t i = 0; i < factors_to_calculate_.size(); i++) {
-            double factor_value = person.get_risk_factor_value(factors_to_calculate_[i]);
-            auto bin_index =
-                static_cast<size_t>((factor_value - factor_min_values_[i]) / factor_bin_widths_[i]);
-            bin_indices.push_back(bin_index);
-        }
+        // First let's fetch the correct `calculated_stats_` bin index for this person
+        size_t index = calculate_index(person);
 
-        // Calculate the index in the calculated_stats_ vector
-        size_t index = 0;
-        for (size_t i = 0; i < bin_indices.size() - 1; i++) {
-            size_t accumulated_bins =
-                std::accumulate(std::next(factor_bins_.cbegin(), i + 1), factor_bins_.cend(),
-                                size_t{1}, std::multiplies<>());
-            index += bin_indices[i] * accumulated_bins * num_factors_to_calculate;
-        }
-        index += bin_indices.back() * num_factors_to_calculate;
-
-        // Now we can add the values of the factors that are not in factors_to_calculate_
-        for (const auto &factor : context.mapping().entries()) {
-            if (std::find(factors_to_calculate_.cbegin(), factors_to_calculate_.cend(),
-                          factor.key()) == factors_to_calculate_.cend()) {
-                calculated_stats_[index++] += person.get_risk_factor_value(factor.key());
+        // Now we can add the calculated stats for this person to the correct index
+        if (!person.is_active()) {
+            if (!person.is_alive() && person.time_of_death() == current_time) {
+                calculated_stats_[index + get_channel_index("deaths")]++;
+                float expected_life =
+                    definition_.life_expectancy().at(context.time_now(), person.gender);
+                double yll = std::max(expected_life - person.age, 0.0f) * DALY_UNITS;
+                calculated_stats_[index + get_channel_index("mean_yll")] += yll;
+                calculated_stats_[index + get_channel_index("mean_daly")] += yll;
             }
+
+            if (person.has_emigrated() && person.time_of_migration() == current_time) {
+                calculated_stats_[index + get_channel_index("emigrations")]++;
+            }
+
+            continue;
+        }
+
+        calculated_stats_[index + get_channel_index("count")]++;
+
+        for (const auto &factor : context.mapping().entries()) {
+            double value = person.get_risk_factor_value(factor.key());
+            calculated_stats_[index + get_channel_index("mean_" + factor.key().to_string())] +=
+                value;
+        }
+
+        for (const auto &[disease_name, disease_state] : person.diseases) {
+            if (disease_state.status == DiseaseStatus::active) {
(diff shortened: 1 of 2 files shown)


src/HealthGPS/analysis_module.cpp
```diff
@@ -319,36 +319,88 @@ DALYsIndicator AnalysisModule::calculate_dalys(Population &population, unsigned
 }
 
 void AnalysisModule::calculate_population_statistics(RuntimeContext &context) {
```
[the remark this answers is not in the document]
alexdewar: I think `clang-tidy` is right that this function is a bit over-complicated... Could you maybe split it up?

src/HealthGPS/analysis_module.cpp
```diff
@@ -560,6 +632,44 @@ void AnalysisModule::initialise_output_channels(RuntimeContext &context) {
...
+    return index;
+}
+
+size_t AnalysisModule::get_channel_index(const std::string &channel) const {
+    auto it = channel_index_.find(channel);
+    if (it == channel_index_.end()) {
+        throw std::out_of_range("Unknown channel: " + channel);
+    }
+
+    return it->second;
```
alexdewar, line 672: Could this be replaced with just `channel_index_.at(channel)`?

src/HealthGPS/analysis_module.cpp
```diff
@@ -531,6 +583,26 @@ void AnalysisModule::classify_weight(DataSeries &series, const Person &entity) c
...
+        calculated_stats_[get_channel_index("over_weight")]++;
+        calculated_stats_[get_channel_index("above_weight")]++;
+        break;
+    case WeightCategory::obese:
+        calculated_stats_[get_channel_index("obese_weight")]++;
+        calculated_stats_[get_channel_index("above_weight")]++;
+        break;
+    default:
+        throw std::logic_error("Unknown weight classification category.");
+        break;
```
alexdewar, line 602: I think you can get away without a `break` after a `throw`:

```suggestion
```

src/HealthGPS/analysis_module.cpp
```diff
@@ -531,6 +583,26 @@ void AnalysisModule::classify_weight(DataSeries &series, const Person &entity) c
...
 
+void AnalysisModule::classify_weight(const Person &person) {
+    auto weight_class = weight_classifier_.classify_weight(person);
+    switch (weight_class) {
+    case WeightCategory::normal:
+        calculated_stats_[get_channel_index("normal_weight")]++;
+        break;
+    case WeightCategory::overweight:
+        calculated_stats_[get_channel_index("over_weight")]++;
+        calculated_stats_[get_channel_index("above_weight")]++;
```
alexdewar, line 594: What's the difference between `over_weight` and `above_weight`?

src/HealthGPS/analysis_module.cpp
```diff
@@ -560,6 +632,44 @@ void AnalysisModule::initialise_output_channels(RuntimeContext &context) {
...
     channels_.emplace_back("std_daly");
+
+    // Since we will be performing frequent lookups, we will store the strings and indexes in a map
+    // for quick access.
+    for (size_t i = 0; i < channels_.size(); i++) {
+        channel_index_.emplace(channels_[i], i);
+    }
+}
+
+size_t AnalysisModule::calculate_index(const Person &person) const {
```
alexdewar, line 643: Could you maybe add a unit test for this function to `HealthGPS.Tests`? I've been pretty bad at writing tests for HealthGPS myself, but this one should be easy enough to write a test for and then we could be a bit more confident it's doing the right thing.

src/HealthGPS/analysis_module.h
```diff
@@ -46,6 +46,7 @@ class AnalysisModule final : public UpdatableModule {
     WeightModel weight_classifier_;
     DoubleAgeGenderTable residual_disability_weight_;
     std::vector<std::string> channels_;
+    std::unordered_map<std::string, int> channel_index_;
```
alexdewar, line 49: I might not be completely understanding what these data structures do... But am I right in thinking `channel_index_` contains indexes into `channels_`? If so, could you just make `channels_` a `std::map<std::string, std::string>` (I'm guessing it needs to be ordered...) and drop `channels_index_`?

alexdewar (changes requested):
Nearly there, I think. I do think it would be worth adding a test for the `calculate_index` function though and I've made a few other small suggestions.

The Windows CI runner is failing because of some type narrowing warning again :disappointed:

src/HealthGPS/analysis_module.cpp
TinyMarsh: I have created an issue addressing this: https://github.com/imperialCHEPI/healthgps/issues/470

src/HealthGPS/analysis_module.cpp
TinyMarsh: Yep! 😅

src/HealthGPS/analysis_module.cpp
TinyMarsh: No idea. They're both in the existing analysis module code though. I guess `above_weight` was added at some point and has some use downstream.

alexdewar (approved):
LGTM!

I'd probs have a go at #470 before merging this, just so we can be confident that the indexing works, but it's up to you. I don't think it needs lots of test cases -- just a few basic ones.

src/HealthGPS/analysis_module.cpp
jamesturner246: One includes everyone in  overweight and obese, and the other just overweight; I forget which is which. TBH the accumulated group is redundant and can be removed, since you can get it by summing the above weight categories anyway.

alexdewar: That's a really confusing naming scheme, though I guess if we change it that will break things :worried:

TinyMarsh:
@jamesturner246 @alexdewar  I'm going to close this PR without merging. The reason is that this body of work is at a point where it's not possible to make iterative progress to the new way of calculating population statistics (essentially by this I mean running the `publish_result_message` method) without breaking what is there currently.

The feedback I have received in this PR has been useful and has been implemented in the `extend_analysis` branch, which I will keep. Further work on the new stats calculation stuff can be merged into that branch. Then, when it is complete, we can merge the `extend_analysis` branch into `main` in one big merge which should implement the new stuff without breaking anything.

I'm aware that our engagement with HealthGPS has essentially ended, but I really think it's worth finishing this since so much work has gone into it and it's quite close to completion now - it shouldn't take a huge amount of effort to get this finished.

PR #482, opened 2024-08-07
pre-commit: Replace prettier hook with separate JSON and YAML formatters

alexdewar:
Sadly, the prettier pre-commit hook seems to be dead. The repo is archived, so we won't be getting updates to the prettier tool.

I've replaced the prettier hook with separate formatters for JSON and YAML, following what we've done in the python-template repo.


alexdewar:
Heh, the JSON formatter appears not to like whatever we've done in `CMakePresets.json`. It looks fine to me though and the `check-json` hook seems to agree it's correct!

I'll dig into it a bit more.

alexdewar:
The problem was that `CMakePresets.json` contained a BOM, which seemed to break the formatter. As BOMs are gross anyway, I've added another hook to remove them.

TinyMarsh (approved):
LGTM! It's subjective but something about putting list items on a separate line when there's only one item in the list irks me. It's whatever though.

alexdewar:
> LGTM! It's subjective but something about putting list items on a separate line when there's only one item in the list irks me. It's whatever though.

Me too :disappointed:. In general I think `prettier`'s formatting is nicer. Oh well.

PR #483, opened 2024-08-07
Refactor code for parsing command-line arguments

alexdewar:
I was trying to change some of the functionality around parsing command-line args and found that the error handling in `parse_arguments` is a bit messy.

There are three cases the function has to handle:

1. The args are valid and the program should start a simulation
2. The args are valid but the program should abort (e.g. if the user passes the `--help` argument)
3. The args are invalid

Previously, the function returned all this information within a `CommandArgs` struct. I've refactored the code so that the function returns a `std::optional<CommandArgs>` instead:

1. If we want to start a simulation, a `CommandArgs` struct is returned
2. If the args are valid, but we want to abort, `std::nullopt` is returned
3. Else an exception is raised

### DIFF
--- a/src/HealthGPS.Console/command_options.cpp
+++ b/src/HealthGPS.Console/command_options.cpp
@@ -19,73 +19,57 @@ cxxopts::Options create_options() {
     return options;
 }
 
-// NOLINTNEXTLINE(modernize-avoid-c-arrays)
-CommandOptions parse_arguments(cxxopts::Options &options, int &argc, char *argv[]) {
+std::optional<CommandOptions> parse_arguments(cxxopts::Options &options, int argc, char **argv) {
     namespace fs = std::filesystem;
 
     CommandOptions cmd;
-    try {
-        cmd.success = true;
-        cmd.exit_code = EXIT_SUCCESS;
-        cmd.verbose = false;
-        auto result = options.parse(argc, argv);
-        if (result.count("help")) {
-            std::cout << options.help() << '\n';
-            cmd.success = false;
-            return cmd;
+    cmd.verbose = false;
+    auto result = options.parse(argc, argv);
+
+    if (result.count("help")) {
+        std::cout << options.help() << '\n';
+        return std::nullopt;
+    }
+
+    if (result.count("version")) {
+        fmt::print("Version {}\n\n", PROJECT_VERSION);
+        return std::nullopt;
+    }
+
+    cmd.verbose = result["verbose"].as<bool>();
+    if (cmd.verbose) {
+        fmt::print(fg(fmt::color::dark_salmon), "Verbose output enabled\n");
+    }
+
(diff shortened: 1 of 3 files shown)


src/HealthGPS.Console/command_options.cpp
```diff
@@ -19,75 +19,57 @@ cxxopts::Options create_options() {
...
-        cmd.success = true;
-        cmd.exit_code = EXIT_SUCCESS;
-        cmd.verbose = false;
-        auto result = options.parse(argc, argv);
-        if (result.count("help")) {
-            std::cout << options.help() << '\n';
-            cmd.success = false;
-            return cmd;
-        }
+    cmd.verbose = false;
```
TinyMarsh, line 26: Is this line necessary because won't `verbose` get value-initialised to `false` anyway?

src/HealthGPS.Console/command_options.cpp
```diff
@@ -19,75 +19,57 @@ cxxopts::Options create_options() {
...
-                       "should pass the data source via the config file.\n");
-            fmt::print("Data source: {}\n", source);
+    if (result.count("storage")) {
+        auto source = result["storage"].as<std::string>();
 
-            cmd.data_source = hgps::input::DataSource(std::move(source));
-        }
+        fmt::print(fmt::fg(fmt::color::yellow),
+                   "WARNING: Path to data source specified with command-line argument. "
+                   "This functionality is deprecatated and will be removed in future. You "
```
TinyMarsh, line 60:
```suggestion
                   "This functionality is deprecated and will be removed in future. You "
```

TinyMarsh (approved):
LGTM. Just a couple things.

src/HealthGPS.Console/command_options.cpp
alexdewar: Nope. The contents of `cmd` are uninitialised at this point, so they could be anything.

If we wrote:

```c++
CommandOptions cmd{};
```

...then it would be. C++ is weird :upside_down_face:

TinyMarsh: Oh right. I thought that when `cmd` got instantiated `verbose` would be value-initialised to `false` because of this:
https://github.com/imperialCHEPI/healthgps/blob/22ad194cfe0e2e62b2f23d92cf9e722d1b379dfa/src/HealthGPS.Console/command_options.h#L24
C++ is weird.

alexdewar: Oh, I didn't notice that! I think you're right. I'll remove it.

alexdewar:
@jamesturner246 I'm heading off now. I'll merge this so that we don't get merge conflicts while I'm gone. The other ones shouldn't cause problems.

jamesturner246:
No worries. Sorry for the delay.

PR #484, opened 2024-08-07
create_intervention_scenario: std::move risk_impacts to avoid copying

alexdewar:
While I was mucking about in the code base, I noticed that we were missing an opportunity to `std::move` something in `create_intervention_scenario()`, so I've fixed it up.


jamesturner246 (approved):
Looks fine. Go ahead :+1:

Issue #486, opened 2024-08-12
Forbid additional properties in data index files


alexdewar:
See: https://github.com/imperialCHEPI/healthgps/pull/404

I initially tried setting `additionalProperties` to false everywhere in the schema, but it broke validation so I didn't bother. I should have opened an issue at the time though, because we do also want to prevent users from adding additional properties (at least most of the time). In part this just makes the files cleaner, but it also prevents users from changing non-existent properties that won't actually do anything (including if they misspell the name of an actual property).

jamesturner246:
> but it also prevents users from changing non-existent properties that won't actually do anything (including if they misspell the name of an actual property).

Precisely the cause of the last fiasco. We probably want to be as pedantic as we can get away with here. :+1:

alexdewar:
Oh dear. Was that with the static data or the config files?

jamesturner246:
That was the static relative disease risk files, in that case.

Issue #452, opened 2024-06-18
Break up `data_index.json` schema file


alexdewar:
This file is currently rather large and bloated. We could break it up into smaller sub-schemas for the sake of readability. @jamesturner246 reports that he is planning to break up the `index.json` files themselves at some point, in which case this change would also make that transition marginally easier.

jamesturner246:
Agreed! :+1:

Issue #173, opened 2023-07-19
Use nholthaus/units for unit safety


jamesturner246:
A number of places --i.e. in risk factor model code -- are quite confusing with respect to the unit that each variable uses.

Using a unit library such as [unit](https://github.com/nholthaus/units) could help clarify and enforce proper unit use.

NOTE: we need to confirm the units used with the modelling team first.

chiphogg:
Stumbled on this issue.  Here's a [GitHub search for C++ units libraries](https://github.com/topics/dimensional-analysis?l=c%2B%2B), and a [comparison of leading options](https://aurora-opensource.github.io/au/main/alternatives/), to help you pick the one that's best for you.  (Disclosure: I'm the author of one of the libraries, the one that is hosting the comparison page.)

alexdewar:
@chiphogg Looks intriguing. Thanks! We'll check it out.

PR #489, opened 2024-08-13
Add JSON schema for top-level config file + use for validation

### LINKED ISSUE #449: Add JSON schema for top-level configuration file
These files are the main config files which provide input data for running user simulations. They (optionally?) include references to config files for static and dynamic risk factor models.

alexdewar:
This PR adds a JSON schema for top-level config files and changes the code to use it for validation. In contrast to the existing schema for `index.json`, I've split the schema into lots of small subschemas (see #452) for readability and maintainability and I've set `additionalProperties` to `false` everywhere (see #486).

There are a couple of small changes that will be needed to existing top-level config files in order for them to work with this new code:

- The schema URL needs to be specified (`"$ref": "https://raw.githubusercontent.com/imperialCHEPI/healthgps/main/schemas/v1/config.json"`)
- The `version` property should be dropped

Other than that, it should work as before. I've tested that all the config files in the examples repo work (or, rather, they will after we merge https://github.com/imperialCHEPI/healthgps-examples/pull/13), but it is possible that the schema is overly restrictive (e.g. a given value can be `null`, but we aren't allowing for this).

I also had to refactor the C++ code for working with schemas to make it more generic, but it essentially does what it did before.

Closes #449.

@jzhu20 FYI. If you send me the config files you're using at present I can check they work with the new schema (hopefully they will!).


schemas/v1/config/interventions.json
```diff
@@ -0,0 +1,104 @@
...
+                    },
+                    "coverage_rates": {
+                        "type": "array",
+                        "items": {
+                            "type": "number"
+                        }
+                    },
+                    "coverage_cutoff_time": {
+                        "type": "integer",
+                        "minValue": 0
```
TinyMarsh, line 70: I don't really know, but it seems the keyword for setting a minimum value is `minimum`, rather than `minValue`. See here:
https://json-schema.org/understanding-json-schema/reference/numeric#range

schemas/v1/config/interventions.json
```diff
@@ -0,0 +1,104 @@
...
+                    },
+                    "child_cutoff_age": {
+                        "type": "integer",
+                        "minValue": 0
```
TinyMarsh, line 74: See above.

schemas/v1/config/modelling.json
```diff
@@ -0,0 +1,112 @@
...
+            "type": "object",
+            "properties": {
+                "function_name": {
+                    "type": "string"
+                },
+                "function_parameters": {
+                    "type": "array",
+                    "items": {
+                        "type": "number"
+                    }
```
TinyMarsh, line 16: Just checking this **array** doesn't require a minimum number of values? Same applies for other arrays in this schema I guess.

schemas/v1/config/modelling.json
TinyMarsh, line 1: Do any of the **string** types here require a `minLength` to ensure they are not empty?

TinyMarsh (approved):
Cool, just a couple of questions.

schemas/v1/config/interventions.json
alexdewar: Good spot! I'm not sure how I managed that :facepalm:. We also have a `pre-commit` hook to check the schemas are valid, but it seems you can add extra properties and it won't complain :shrug:

schemas/v1/config/modelling.json
alexdewar: I'm not sure... @jamesturner246?

schemas/v1/config/modelling.json
alexdewar: That's a good idea. I don't know whether I can be bothered to put `minLength`s all over the place though, but maybe I'm just being lazy :laughing:.

What do you think @jamesturner246? How pedantic do we want to be?

alexdewar:
Thanks for going through this @TinyMarsh. I know it's probably not much fun just reviewing a bunch of JSON :laughing:

schemas/v1/config/modelling.json
jamesturner246: I won't push you -- one would hope users know not to put empty string fields... Certainly won't hurt though.

schemas/v1/config/modelling.json
jamesturner246: I think these depend on the SES model -- e.g. normal has mean, std. Maybe there are other models that have none? Never uses the SES module.

src/HealthGPS.Input/schema.cpp
```diff
@@ -26,23 +32,63 @@ json resolve_uri(const jsoncons::uri &uri, const std::filesystem::path &schema_d
...
-    // **YUCK**: We have to read in the data with jsoncons here rather than reusing
-    // the nlohmann-json representation :-(
-    const auto index = json::parse(index_stream);
+/// @brief Validate a JSON file against the specified schema
+/// @param json_stream The input stream for the JSON file
+/// @param schema_file_name The name of the JSON schema file
+/// @param schema_version The version of the schema file
+void validate_json(std::istream &json_stream, const char *schema_file_name, int schema_version) {
+    // **YUCK**: We have to read in the data with jsoncons here rather than reusing the
+    // nlohmann-json representation :-(
```
jamesturner246, line 41: Would there be any benefit to switching over to another JSON lib, e.g. jsoncons in future? I saw an issue opened by someone: https://github.com/imperialCHEPI/healthgps/issues/173#issuecomment-2191850590. No need to worry now.

jamesturner246 (approved):
LGTM! 

I won't push to add extra validation like string length checking now if it's tons of work. This looks like the bulk of it. And some of the arrays like SES module I don't even know how long they are supposed to be. We can look at it later if problems occur.

schemas/v1/config/modelling.json
alexdewar: Hmm... ok. I'll leave as is for now. We can always add extra constraints in future if needed.

src/HealthGPS.Input/schema.cpp
alexdewar: I'm not sure. `nlohmann-json` is kind of the standard one, but it seems like `jsoncons` has a very similar interface. It bugs me that we're using two JSON libraries, but I fear it might involve a fair amount of refactoring to swap everything over to `jsonconss`.

alexdewar:
After our discussion in the meeting, I think I'll change this so that it's fully backwards compatible with existing config files, just to make everyone's life easier. Shouldn't be hard to do.

PR #491, opened 2024-08-14
Time trend for expected values

### LINKED ISSUE #477: Add time trend in risk factors
We are adding an underlying time trend in risk factors from consuming UPF (ultra-processed food).
The plan is to add time (or year) as an additional variable into risk factor equations.

- [x] The parameters need to be estimated.
- [x] It needs to be confirmed how this time trend will affect other equations in Health-GPS.
- [x] #492 
- [x] #493

### LINKED ISSUE #492: Add time trend to expected RF values
Expected values trend over time. We need to add and apply exponential time factors to the returned expected values in `RiskFactorAdjustableModel`.

jamesturner246:
Resolves #492 

This change adds new `StaticLinear.json` params: `ExpectedTrend`, which are raised to the power of the iteration number and used to multiply risk factor expected values, to simulate (exponentially) changing population average of nutrients over time.

These are passed into adjustable models as shared pointers rather than references -- see https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/avoid-const-or-ref-data-members.html

All adjustable RF model code is updated to use the new trends.

### DIFF
--- a/src/HealthGPS.Input/model_parser.cpp
+++ b/src/HealthGPS.Input/model_parser.cpp
@@ -332,9 +332,9 @@ load_staticlinear_risk_model_definition(const nlohmann::json &opt, const Configu
     const double physical_activity_stddev = opt["PhysicalActivityStdDev"].get<double>();
 
     return std::make_unique<hgps::StaticLinearModelDefinition>(
-        std::move(expected), std::move(names), std::move(models), std::move(ranges),
-        std::move(lambda), std::move(stddev), std::move(cholesky), std::move(policy_models),
-        std::move(policy_ranges), std::move(policy_cholesky), info_speed,
+        std::move(expected), std::move(expected_trend), std::move(names), std::move(models),
+        std::move(ranges), std::move(lambda), std::move(stddev), std::move(cholesky),
+        std::move(policy_models), std::move(policy_ranges), std::move(policy_cholesky), info_speed,
         std::move(rural_prevalence), std::move(income_models), physical_activity_stddev);
 }
 
--- a/src/HealthGPS.Input/model_parser.cpp
+++ b/src/HealthGPS.Input/model_parser.cpp
@@ -178,14 +178,14 @@ load_staticlinear_risk_model_definition(const nlohmann::json &opt, const Configu
                                       policy_covariance_table.num_columns()};
 
     // Risk factor and intervention policy: names, models, parameters and correlation/covariance.
-    std::vector<hgps::core::Identifier> names;
-    std::vector<hgps::LinearModelParams> models;
-    std::vector<hgps::core::DoubleInterval> ranges;
+    std::vector<core::Identifier> names;
+    std::vector<LinearModelParams> models;
+    std::vector<core::DoubleInterval> ranges;
     std::vector<double> lambda;
     std::vector<double> stddev;
-    std::vector<hgps::LinearModelParams> policy_models;
-    std::vector<hgps::core::DoubleInterval> policy_ranges;
-    std::vector<double> expected_trend;
+    std::vector<LinearModelParams> policy_models;
+    std::vector<core::DoubleInterval> policy_ranges;
+    std::unordered_map<core::Identifier, double> expected_trend;
 
     size_t i = 0;
     for (const auto &[key, json_params] : opt["RiskFactorModels"].items()) {
@@ -243,7 +243,7 @@ load_staticlinear_risk_model_definition(const nlohmann::json &opt, const Configu
         }
 
         // Load expected value trends.
-        expected_trend.emplace_back(json_params["ExpectedTrend"].get<double>());
+        expected_trend[key] = json_params["ExpectedTrend"].get<double>();
 
         // Increment table column index.
         i++;
--- a/src/HealthGPS.Input/model_parser.cpp
+++ b/src/HealthGPS.Input/model_parser.cpp
@@ -24,7 +24,8 @@
 
 namespace hgps::input {
 
-hgps::RiskFactorSexAgeTable load_risk_factor_expected(const Configuration &config) {
+std::unique_ptr<hgps::RiskFactorSexAgeTable>
+load_risk_factor_expected(const Configuration &config) {
     MEASURE_FUNCTION();
 
     const auto &info = config.modelling.baseline_adjustment;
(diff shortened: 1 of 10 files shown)


alexdewar:
@jamesturner246 I'm guessing you meant to tag us as reviewers rather than assigning this to us?

jamesturner246:
Oh whoops. Thanks!

src/HealthGPS.Input/model_parser.cpp
```diff
@@ -24,29 +24,30 @@
...
+load_risk_factor_expected(const Configuration &config) {
     MEASURE_FUNCTION();
 
     const auto &info = config.modelling.baseline_adjustment;
     if (!hgps::core::case_insensitive::equals(info.format, "CSV")) {
         throw hgps::core::HgpsException{"Unsupported file format: " + info.format};
     }
 
-    auto table = hgps::RiskFactorSexAgeTable{};
+    auto table = std::make_unique<hgps::RiskFactorSexAgeTable>();
```
alexdewar, line 36: Why did you change `table` to a `std::unique_ptr` btw? Not a problem; just curious.

src/HealthGPS.Input/model_parser.cpp
```diff
@@ -268,15 +273,15 @@ load_staticlinear_risk_model_definition(const nlohmann::json &opt, const Configu
         Eigen::MatrixXd{Eigen::LLT<Eigen::MatrixXd>{policy_covariance}.matrixL()};
 
     // Risk factor expected values by sex and age.
-    hgps::RiskFactorSexAgeTable expected = load_risk_factor_expected(config);
+    std::unique_ptr<hgps::RiskFactorSexAgeTable> expected = load_risk_factor_expected(config);
```
alexdewar, line 276: I'd generally use `auto` in place of longer type names:

```suggestion
    auto expected = load_risk_factor_expected(config);
```

But up to you.

src/HealthGPS.Input/model_parser.cpp
```diff
@@ -425,10 +430,12 @@ load_ebhlm_risk_model_definition(const nlohmann::json &opt, const Configuration
     }
 
     // Risk factor expected values by sex and age.
-    hgps::RiskFactorSexAgeTable expected = load_risk_factor_expected(config);
+    std::unique_ptr<hgps::RiskFactorSexAgeTable> expected = load_risk_factor_expected(config);
```
alexdewar, line 433: Ditto:

```suggestion
    auto expected = load_risk_factor_expected(config);
```

src/HealthGPS.Input/model_parser.cpp
```diff
@@ -437,7 +444,8 @@ load_kevinhall_risk_model_definition(const nlohmann::json &opt, const Configurat
     MEASURE_FUNCTION();
 
     // Risk factor expected values by sex and age.
-    hgps::RiskFactorSexAgeTable expected = load_risk_factor_expected(config);
+    std::unique_ptr<hgps::RiskFactorSexAgeTable> expected = load_risk_factor_expected(config);
```
alexdewar, line 447:
```suggestion
    auto expected = load_risk_factor_expected(config);
```

src/HealthGPS.Input/model_parser.cpp
```diff
@@ -24,29 +24,30 @@
 
 namespace hgps::input {
 
-hgps::RiskFactorSexAgeTable load_risk_factor_expected(const Configuration &config) {
+std::unique_ptr<hgps::RiskFactorSexAgeTable>
+load_risk_factor_expected(const Configuration &config) {
```
alexdewar, line 28: I suppose this *could* be a member function...

src/HealthGPS/risk_factor_adjustable_model.cpp
```diff
@@ -31,11 +32,18 @@ struct FirstMoment {
...
+    std::shared_ptr<RiskFactorSexAgeTable> expected,
+    std::shared_ptr<std::unordered_map<core::Identifier, double>> expected_trend)
+    : expected_{std::move(expected)}, expected_trend_{std::move(expected_trend)} {}
+
+double RiskFactorAdjustableModel::get_expected(RuntimeContext &context, core::Gender sex, int age,
+                                               const core::Identifier &factor) const noexcept {
+    double expected = expected_->at(sex, factor).at(age);
+    if (expected_trend_->contains(factor)) {
+        int elapsed_time = context.time_now() - context.start_time();
+        expected *= pow(expected_trend_->at(factor), elapsed_time);
```
alexdewar, line 44: Are these arguments the wrong way round? I'm trying to wrap my head round why you'd raise to the power of `elapsed_time`.

alexdewar (approved):
Besides my question about the use of `pow()`, it all looks fine to me. I've made a few small suggestions.

src/HealthGPS/risk_factor_adjustable_model.cpp
jamesturner246: Yeah, it' look a bit strange, but it's basically an exponential trend in time, With time factor `t`, we have `with_trend = without_trend * t^elapsed`. So for year zero, we have `with_trend = without_trend * t^0 = without_trend`, and `t` factor changes exponentially with years.

src/HealthGPS.Input/model_parser.cpp
jamesturner246: I noted clang-tidy began complaining about model members as const and reference complicating automatic ctor generation -- see https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/avoid-const-or-ref-data-members.htmlv and links therein.

I suppose I could have made it into a shared pointer within the model class, but thought this way we only std move a pointer to the definition ctor. :shrug:

src/HealthGPS.Input/model_parser.cpp
jamesturner246: Of the `AdjustableRiskFactorModelDefinition` class, you mean?

src/HealthGPS.Input/model_parser.cpp
alexdewar: Ah ok. That's interesting.

I think it's the right decision to return it as a `std::unique_ptr`. Out of interest, is it actually shared though? Or could it be left as a `std::unique_ptr`?

src/HealthGPS/risk_factor_adjustable_model.cpp
jamesturner246: That reminds me, I need to bound these values...

src/HealthGPS.Input/model_parser.cpp
jamesturner246: It's shared once it enters the risk factor model definition class (with the risk factor model class). I keep it unique here and move it into a shared pointer inside the definition class. Not sure if that's good practice or not..

PR #496, opened 2024-08-16
Add some basic VS Code settings

alexdewar:
It periodically annoys me when I find that files aren't being autoformatted in VS Code or the rulers are in the wrong place for HGPS. Let's add these settings for VS Code users.

I've also added the C++ extension pack (which includes CMake support) and a spellchecker to the extension recommendations.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,19 @@
+{
+    "[cpp]": {
+        "editor.formatOnSave": true,
+        "editor.rulers": [100],
+    },
+    "[cmake]": {
+        "editor.formatOnSave": true,
+        "editor.rulers": [100],
+    },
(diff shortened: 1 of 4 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,19 @@
...
+    },
+    "[json]": {
+        "editor.formatOnSave": false
+    },
+    "[jsonc]": {
+        "editor.formatOnSave": false
+    },
+    "[yaml]": {
+        "editor.formatOnSave": false
+    }
```
jamesturner246, line 18: Home come the config file rules are disabled?

jamesturner246 (approved):
LGTM :+1:  Just a question.

PR #498, opened 2024-08-20
Time trend policy fix

### LINKED ISSUE #497: policy should be a fn of non-trend risk factors
It turns out that intervention policies should be computed from risk factors without time trend applied. They are currently computed from factors with trend applied.

Fix to use non-trend factors.

jamesturner246:
Resolves #497 

We want policy updates to be computed as a function of *non-time-trended, expected val adjusted* nutrient values. This means we need to do a big ugly second loop for expected value adjustment: the first without time trend, and the second one (after policy is computed) to produce the final value.

For this, we add a boolean flag `apply_trend` to the expect adjustment methods. We also separate the policy updating into separate compute and apply step functions, so that we can compute policies on non trended nutrients, and apply policies after time trends are applied.


alexdewar (approved):
I've only given it a fairly superficial look, but it seems to make sense.

PR #503, opened 2024-08-27
Split up `data_index.json` schema file

### LINKED ISSUE #452: Break up `data_index.json` schema file

alexdewar:
No functional change intended. I have tested that the existing data is loadable after this change.

Closes #452.


jamesturner246 (approved):
LGTM. Much nicer.

PR #506, opened 2024-08-28
Fix rebasing of relative data paths

alexdewar:
Users are able to specify a path to a zip file or directory via the config file, as well as a URL. If the path is a relative path, we treat it as relative to the folder the config file is in, but currently this doesn't work.

The logic of `try_rebase_path()` is currently broken, as it returns early if `std::filesystem::is_directory()` returns false. There are a couple of things wrong with this:

1. If it is a relative path, then false will be returned anyway, because non-existent paths also cause false to be returned
2. We don't actually need to distinguish between whether the path is to a zip file or a directory, just whether or not it is a URL

### DIFF
--- a/src/HealthGPS.Core/math_util.cpp
+++ b/src/HealthGPS.Core/math_util.cpp
@@ -1,4 +1,6 @@
 #include "math_util.h"
+
+#include <algorithm>
 #include <cmath>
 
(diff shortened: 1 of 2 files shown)


jamesturner246 (approved):
Looks good. :+1:

PR #507, opened 2024-08-28
CMake: Make VS Code use same formatter as we use with `pre-commit`

alexdewar:
I changed the VS Code settings to auto-format CMake files, but unfortunately the default formatter doesn't use the same `cmake-format` tools as the `pre-commit` hook uses, so they're fighting one another. Fix this by changing the VS Code settings to use a compatible extension for formatting CMake files.


jamesturner246 (approved):
LGTM :+1:

PR #508, opened 2024-08-28
Fix CI coverage

### LINKED ISSUE #505: [BUG] Coverage no longer displaying in README badge
I'm not sure when this bug manifested, but we should fix it. There's no point in having the badge if it's not working. This may also mean that it's not calculating changes to coverage for PRs, which is something we want.

alexdewar:
Fixes #505.


alexdewar:
I managed to fix this without actually changing the code.

PR #509, opened 2024-08-29
Disallow additional properties for data index files

### LINKED ISSUE #486: Forbid additional properties in data index files

alexdewar:
This PR fixes up the schema for data index files and the associated subschemas so that additional properties are disallowed (i.e. if a user includes a property in their file called `my_madeup_property` they will now get an error when loading the data).

This can catch a number of problems:

1. A user thinks a property exists when it doesn't and doesn't notice that changing its value does nothing
2. A user misspells the name of an optional property and they end up with a default value for the simulation without knowing about it
3. Properties that we've forgotten to include in schemas (I fixed one such bug on this branch)

It seems that combining `allOf` with `additionalProperties==false` doesn't work, unfortunately, so I've essentially ended up copying the contents of the `file_info.json` subschema all over the place (and deleting it). We wouldn't have this problem if all the file info-related properties were housed in single properties as opposed to being mixed in with other values, so maybe we should make this the policy for v2 of the data index schema.

Closes #486.


jamesturner246 (approved):
I see. Guess it is what it is. Agree on the file schema thing for v2. Considering the state it was before, these incremental improvements are steps towards the eventual perfect schema. LGTM :+1: 

Question: what was `allOf` doing here?

alexdewar:
The `allOf` was there to say "you need all of the properties defined in `file_info.json` plus these other ones".

PR #513, opened 2024-09-03
Allow for specifying output folder via command-line argument

### LINKED ISSUE #512: Change output folder from config file option to command-line argument

Currently users have to set the desired output directory for results via a config file option, which isn't great for a number of reasons:

1. Users may share a given config file between different machines (or with colleagues) where a different output path would make sense
2. It is easy to forget to change this value
3. Really, it's a program setting rather than a model parameter and it's odd and confusing to mix the two

@jzhu20 has confirmed that she finds this annoying :stuck_out_tongue: 

We should add a command-line option to set the output path. As with setting the data directory, we should deprecate setting the output path via the config file and display a warning to users who do so, but not break existing config files.

alexdewar:
This PR adds the option to specify the output folder via a command-line argument instead of through an option in a config file. The latter is still supported, but a warning will be displayed, as longer term it would be good to move users away from this. Note that users still need to specify a file name in the config file for now, as this option is actually used to create two file names (for the JSON and CSV output files). (Plus, the name can also include a placeholder string representing a time stamp and it seemed weird to include this in the command-line argument.)

Closes #512.

### DIFF
--- a/schemas/v1/config/output.json
+++ b/schemas/v1/config/output.json
@@ -15,7 +15,6 @@
     },
     "required": [
         "comorbidities",
-        "folder",
         "file_name"
     ],
     "additionalProperties": false
--- a/src/HealthGPS.Console/command_options.cpp
+++ b/src/HealthGPS.Console/command_options.cpp
@@ -11,6 +11,7 @@ cxxopts::Options create_options() {
     cxxopts::Options options("HealthGPS.Console", "Health-GPS microsimulation for policy options.");
     options.add_options()("f,file", "Configuration file full name.", cxxopts::value<std::string>())(
         "s,storage", "Path to root folder of the data storage.", cxxopts::value<std::string>())(
+        "o,output", "Path to output folder", cxxopts::value<std::string>())(
         "j,jobid", "The batch execution job identifier.",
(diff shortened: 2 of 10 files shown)


schemas/v1/config/output.json
```diff
@@ -15,7 +15,6 @@
     },
     "required": [
         "comorbidities",
-        "folder",
```
jamesturner246, line 18: Won't this break back compatibility with `"additionalProperties": false`?

jamesturner246 (approved):
Looks good. One question though.

Also, regarding the timestamp in cmd option thing, maybe we can allow users to pass `... -o ./my/dir/<some syntax indicating a "variable">`, and allow timestamps that way, if that's what they want.

Maybe even just `TIMESTAMP` all caps would suffice.

schemas/v1/config/output.json
jamesturner246: Oh no, I see, `folder` is still listed above, just not required.

alexdewar:
I'm humming and hawing about whether to let users specify the full path for output files via the command-line argument, given some subtleties about the whole `TIMESTAMP` placeholder thing, so I've just opened an issue about it for now: #516 

In the meantime, let's just merge this as an incremental improvement.

PR #515, opened 2024-09-11
Set max threads with command-line argument

alexdewar:
The current method for setting the maximum number of threads is to set the environment variable OMP_THREAD_LIMIT. While HealthGPS doesn't (presently) use OpenMP, this was introduced for the sake of backwards compatibility; the older version of HealthGPS *did* use OpenMP.

The problem is that OMP_THEAD_LIMIT isn't the only environment variable used to control the number of threads with OpenMP: the usual one to set is OMP_NUM_THREADS. Unfortunately, Imperial's HPC cluster doesn't automatically set OMP_THREAD_LIMIT, just OMP_NUM_THREADS, meaning that the current hack won't actually work: users have to explicitly set OMP_THREAD_LIMIT to control parallelism.

I think it would be cleaner to abandon the hack altogether and let users (optionally) set the parallelism via a command-line argument, which is what I've implemented here.

As this is technically a breaking change, I've opened the PR in draft.

### DIFF
--- a/src/HealthGPS.Console/command_options.cpp
+++ b/src/HealthGPS.Console/command_options.cpp
@@ -11,10 +11,13 @@ cxxopts::Options create_options() {
     cxxopts::Options options("HealthGPS.Console", "Health-GPS microsimulation for policy options.");
     options.add_options()("f,file", "Configuration file full name.", cxxopts::value<std::string>())(
         "s,storage", "Path to root folder of the data storage.", cxxopts::value<std::string>())(
-        "j,jobid", "The batch execution job identifier.",
-        cxxopts::value<int>())("verbose", "Print more information about progress",
-                               cxxopts::value<bool>()->default_value("false"))(
-        "help", "Help about this application.")("version", "Print the application version number.");
+        "j,jobid", "The batch execution job identifier.", cxxopts::value<int>())(
+        "T,threads", "The maximum number of threads to create (0: no limit, default).",
+        cxxopts::value<size_t>())
+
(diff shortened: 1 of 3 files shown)


src/HealthGPS.Console/command_options.cpp
```diff
@@ -11,10 +11,13 @@ cxxopts::Options create_options() {
     cxxopts::Options options("HealthGPS.Console", "Health-GPS microsimulation for policy options.");
     options.add_options()("f,file", "Configuration file full name.", cxxopts::value<std::string>())(
         "s,storage", "Path to root folder of the data storage.", cxxopts::value<std::string>())(
-        "j,jobid", "The batch execution job identifier.",
-        cxxopts::value<int>())("verbose", "Print more information about progress",
-                               cxxopts::value<bool>()->default_value("false"))(
-        "help", "Help about this application.")("version", "Print the application version number.");
+        "j,jobid", "The batch execution job identifier.", cxxopts::value<int>())(
+        "T,threads", "The maximum number of threads to create (0: no limit, default).",
```
jamesturner246, line 15: Why uppercase here?

jamesturner246 (approved):
Looks like it does the job. One nitpick.

@jzhu20 will need to be aware of this when merged.

src/HealthGPS.Console/command_options.cpp
alexdewar: It seems to be standard for a lot of tools (e.g. `xz`) but I'm not sure why! It can be lowercase if you prefer.

PR #519, opened 2024-09-11
Allow users to load models from URLs and zip files

### LINKED ISSUE #406: Allow for running example from URL
Depends on #364.

I think this would be a useful QoL improvement for users, because they could then run examples (or other users' models) from a URL directly without having to download it. We could reuse the infrastructure for data input files.

What do you think @jamesturner246?

### LINKED ISSUE #490: Standardise name of top-level config files
At present, users can choose whatever name they like for the config file (though generally we are using names ending in `*Config.json`). The whole path to this config file is passed to HealthGPS when we start it. This is in contrast to the top-level JSON file for static data, which is always called `index.json`.

The problem is that if we want to implement #406, there will be no way for HGPS to know what the name of the top-level config file is inside a zip file. We could also require users to pass in the filename separately, but this seems a bit gross.

It might be best just to standardise the filename (e.g. to `config.json`). We could do this in a backwards compatible way: if users pass in the path to a JSON file that is called something else, we could just give them a deprecation warning.

Am I missing anything @jamesturner24? One downside of this approach is that it will no longer be possible to have multiple top-level config files in the same directory (well, without using deprecated functionality, anyway).

alexdewar:
This PR adds the option to load a config directory from a zip file, either on the local file system or at a given URL, similar to what we do with the static data currently. This is mostly a convenience, but it should help us make proper releases for the config files in `healthgps-examples`, because users will now run a given example using just the zip file or the URL to it. If users want to load a configuration from a local directory that's still totally fine too though! There's no plan to deprecate this functionality (in contrast to the static data).

I split out the checksum checking from the `DataSource` class into a separate `ValidatedDataSource` class which I think is a cleaner design, though I could have left it as it was.

Tagging @jzhu20 FYI.

Closes #406. Closes #490.


src/HealthGPS.Input/data_source.cpp
```diff
@@ -36,73 +20,50 @@ std::filesystem::path get_zip_file_path(const std::string &source) {
...
-DataSource::DataSource(std::string source, const std::filesystem::path &root_path,
-                       std::optional<std::string> file_hash)
-    : source_(try_rebase_path(std::move(source), root_path)), file_hash_(std::move(file_hash)),
-      validate_checksum_(true) {}
+DataSource::DataSource(std::string source, const std::filesystem::path &root_path)
+    : source_(try_rebase_path(std::move(source), root_path)) {}
+
+std::string DataSource::compute_file_hash(const std::filesystem::path &zip_file_path) const {
+    return hgps::compute_sha256_for_file(zip_file_path);
+}
```
jamesturner246, line 48: Is this unvalidated hash used in caching? Else not sure why it's needed in unvalidated datasource.

src/HealthGPS.Input/validated_data_source.cpp
jamesturner246, line 1: So this is identical to `DataSource`, except the overloaded `compute_hash` also validates a given hash?

jamesturner246 (approved):
Approve, but it's getting a bit harder to follow now, with all the compatibility mingled in. 

We should talk about compatibility layers soon, to keep mainline simple. As in a set of legacy wrapper functions (handling all the corner cases) which call the underlying clean modernised interface.

src/HealthGPS.Console/command_options.cpp
```diff
@@ -41,16 +43,15 @@ std::optional<CommandOptions> parse_arguments(cxxopts::Options &options, int arg
...
     }
 
-    cmd.config_file = result["file"].as<std::string>();
-    if (cmd.config_file.is_relative()) {
-        cmd.config_file = std::filesystem::absolute(cmd.config_file);
-        fmt::print("Configuration file: {}\n", cmd.config_file.string());
+    if (result.count("file")) {
+        fmt::print(fg(fmt::color::dark_salmon),
+                   "The -f/--file option is deprecated. Use -c/--config instead.\n");
+        cmd.config_source = result["file"].as<std::string>();
```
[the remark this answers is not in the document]
alexdewar: Yep!

src/HealthGPS.Input/data_source.cpp
alexdewar: Yep, that's it.

src/HealthGPS.Input/validated_data_source.cpp
alexdewar: Yup!

Issue #521, opened 2024-09-12
[BUG] `KevinHall_India` example doesn't work with latest commit of HealthGPS


alexdewar:
## Describe the bug

The version of HealthGPS on `main` no longer loads the `KevinHall_India` example, since #491. I think this is because of the new `ExpectedTrend` property, which doesn't exist in the example.

## To Reproduce

Run `HealthGPS.Console` with `heathgps-examples/KevinHall_India/model/Config.json`.

## Expected behavior

For the simulation to run, as it did before #491.

## Software versions

- OS: Linux
- Health-GPS commit hash: 25a516f344851b3af69f99387ebf1c4cfdd3c8fe

## Additional context

As I see it, there are two ways we can fix this:

1. Add the missing property/properties to the example
2. Change HealthGPS to fallback on the previous behaviour if the property isn't present

This is also probably a good illustration of why it would be handy to make regular releases for both `healthgps` and `healthgps-examples` (I appreciate this is hard when development is so constant though!).

Provided this problem isn't affecting @jzhu20, it's not that big of a deal, but it would be useful to fix it at some point when I come to writing JSON schemas for the model files.

Any thoughts @jamesturner246?

alexdewar:
Superseded by https://github.com/imperialCHEPI/healthgps-examples/issues/28 and #531.

Issue #487, opened 2024-08-12
Config files can contain multiple seeds, even though only one is used


alexdewar:
See: https://github.com/imperialCHEPI/healthgps/blob/4cfa3c271bad3bfb956fdadd0d2bc96af8e7a217/src/HealthGPS.Input/configuration_parsing.cpp#L238

Note that although `seed` is an array, we only ever read at most one value from it. If the array is empty or the property is missing altogether, the seed is automatically generated.

A better way to express this would be to make it so the seed can be either an integer or `null`. Not sure whether the property should be required or optional.

jamesturner246:
That is weird. You say only the first seed is used and the rest discarded? Gross. Integer or null is good.

I'd personally make everything required and disable with null, but that's just me. Happy with what you choose.

alexdewar:
> That is weird. You say only the first seed is used and the rest discarded? Gross. Integer or null is good.

Yeah, exactly.

> I'd personally make everything required and disable with null, but that's just me. Happy with what you choose.

I think that's the right call too. Better to be explicit.

PR #525, opened 2024-09-18
Add schemas for model configs

### LINKED ISSUE #450: Add JSON schemas for static risk factor models
While there are multiple options for static risk factor models, some of the options will be common and these parts can be pulled out into a separate schema. Individual models will require their own schemas for model-specific parameters.

### LINKED ISSUE #451: Add JSON schemas for dynamic risk factor models
This is a very similar problem to defining schemas for the static risk factor models. See #450.

alexdewar:
This PR adds JSON schemas for both static and dynamic model configuration files. I've used the files in `healthgps-examples` as a reference, which means that the schemas are currently slightly out of step with the model parsing code in HealthGPS (see #521). I don't intend to enforce these schemas for now though -- I just want them as a reference for when we make a release for `healthgps-examples`. Once we've done that, I'll start on some v2 schemas that we can actually enforce (though I'll probably still need some example config files to work from).

All the config files in `healthgps-examples` validate successfully with these schemas :partying_face:.

There is one small unrelated fix: I noticed that the schema for the `index.json` file doesn't allow the user to omit the random seed (which they can do by providing an empty array: #487) so I've changed that.

Closes #450. Closes #451.

### DIFF
--- a/schemas/v1/config/csv_file.json
+++ b/schemas/v1/config/csv_file.json
@@ -1,0 +1,40 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "https://raw.githubusercontent.com/imperialCHEPI/healthgps/main/schemas/v1/config/csv_file.json",
+    "type": "object",
+    "properties": {
+        "name": {
+            "type": "string"
+        },
+        "format": {
+            "const": "csv"
+        },
+        "delimiter": {
+            "const": ","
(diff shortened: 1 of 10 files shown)


jamesturner246 (approved):
Looks good. Probably for the best not to enforce these just yet, like you say. The latest model changes add yet more fields, but they can be done in another PR. :+1:

PR #527, opened 2024-09-19
Child weight in KevinHall::get_expected.

### LINKED ISSUE #522: [BUG] Children's BMI extremely high
## Describe the bug

After adding time trend, children's BMI is extremely high.

## To Reproduce

1. Using config files '...'
2. Use command '...'
3. See issue '...' inside output files/text '...'

## Expected behavior

Children's BMI within a normal range. After introducing time trend to nutrients and energy intake, children's expected weight should be adjusted with a different equation from adults.
The correct equation is
`expected_weight_adjusted = weight_in_factors_mean * (expected_energy_intake_with_time_trend / expected_energy_intake_in_factors_mean)^0.45`

## Software versions

- OS: HPC
- Health-GPS commit hash 282a6140f9c6b6cdf6e83a9a09de28bccadd4ed0

## Additional context

NA

jamesturner246:
This change modifies `KevinHallModel::get_expected` to handle children differently. This is because we are using separate KevinHall models for both children and adults, and the expected value calculation needs to account for this.

Resolves #522

### DIFF
--- a/src/HealthGPS/kevin_hall_model.cpp
+++ b/src/HealthGPS/kevin_hall_model.cpp
@@ -200,10 +200,10 @@ void KevinHallModel::update_non_newborns(RuntimeContext &context) const {
     // Compute weight power means by sex and age.
     auto W_power_means = compute_mean_weight(context.population(), height_slope_);
 
-    // Update: (no newborns or 19 and over).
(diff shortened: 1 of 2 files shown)


jzhu20 (commented):
It looks good to me. The equations are correct.

alexdewar (approved):
LGTM :+1:

PR #528, opened 2024-09-19
Trend max steps

### LINKED ISSUE #517: Fix time coefficient after 20 years

We are using an exponential time coefficient to allow time trend in nutrient intake. This time coefficient needs to be fixed after 20 years of simulation.

Keep the exponential time coefficient for the first 20 years, and keep it constant from the 21st year onwards.
Based on my understanding of the code, when updating time trend, we can add a range of not exceeding 20 to elapsed_time.
https://github.com/imperialCHEPI/healthgps/blob/main/src/HealthGPS/static_linear_model.cpp#L209-L210
Then the time coefficient will be fixed after 20 years.
The same applies when expected nutrients are adjusted with time trend. https://github.com/imperialCHEPI/healthgps/blob/main/src/HealthGPS/risk_factor_adjustable_model.cpp#L45-L48
The elapsed_time can be restricted to not over 20.

## Describe alternatives you've considered

NA

## Additional context

NA

jamesturner246:
*REVIEW AFTER #527*

This change adds a new configurable `TrendSteps` value (must be defined for both the static (i.e. `StaticLinear.json`) and the dynamic (i.e. `KevinHall.json`) models).

This is used to ensure that the exponential trends in `RiskFactorAdjustableModel` and `StaticLinearModel`are capped after a defined number of elapsed time steps.

Resolves #517

### DIFF
--- a/src/HealthGPS/risk_factor_adjustable_model.cpp
+++ b/src/HealthGPS/risk_factor_adjustable_model.cpp
@@ -114,6 +114,10 @@ void RiskFactorAdjustableModel::adjust_risk_factors(RuntimeContext &context,
     }
 }
 
+int RiskFactorAdjustableModel::get_trend_steps(const core::Identifier &factor) const noexcept {
+    return trend_steps_->at(factor);
+}
+
 RiskFactorSexAgeTable
 RiskFactorAdjustableModel::calculate_adjustments(RuntimeContext &context,
                                                  const std::vector<core::Identifier> &factors,
--- a/src/HealthGPS/risk_factor_adjustable_model.cpp
+++ b/src/HealthGPS/risk_factor_adjustable_model.cpp
@@ -33,8 +33,10 @@ namespace hgps {
 
 RiskFactorAdjustableModel::RiskFactorAdjustableModel(
     std::shared_ptr<RiskFactorSexAgeTable> expected,
-    std::shared_ptr<std::unordered_map<core::Identifier, double>> expected_trend)
-    : expected_{std::move(expected)}, expected_trend_{std::move(expected_trend)} {}
+    std::shared_ptr<std::unordered_map<core::Identifier, double>> expected_trend,
+    std::shared_ptr<std::unordered_map<core::Identifier, int>> trend_steps)
+    : expected_{std::move(expected)}, expected_trend_{std::move(expected_trend)},
+      trend_steps_{trend_steps} {}
 
 double RiskFactorAdjustableModel::get_expected(RuntimeContext &context, core::Gender sex, int age,
                                                const core::Identifier &factor, OptionalRange range,
--- a/src/HealthGPS/risk_factor_adjustable_model.cpp
+++ b/src/HealthGPS/risk_factor_adjustable_model.cpp
@@ -193,14 +193,19 @@ RiskFactorAdjustableModel::calculate_simulated_mean(Population &population,
 
 RiskFactorAdjustableModelDefinition::RiskFactorAdjustableModelDefinition(
     std::unique_ptr<RiskFactorSexAgeTable> expected,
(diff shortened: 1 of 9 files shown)


jzhu20:
It looks good to me.

src/HealthGPS/risk_factor_adjustable_model.cpp
```diff
@@ -44,7 +46,8 @@ double RiskFactorAdjustableModel::get_expected(RuntimeContext &context, core::Ge
     // Apply optional trend to expected value.
     if (apply_trend) {
         int elapsed_time = context.time_now() - context.start_time();
-        expected *= pow(expected_trend_->at(factor), elapsed_time);
+        int t = std::min(elapsed_time, get_trend_steps(factor));
+        expected *= pow(expected_trend_->at(factor), t);
```
alexdewar, line 50: Could you explain the logic of this bit? It's puzzling me.

src/HealthGPS/risk_factor_adjustable_model.cpp
```diff
@@ -193,15 +200,20 @@ RiskFactorAdjustableModel::calculate_simulated_mean(Population &population,
 
 RiskFactorAdjustableModelDefinition::RiskFactorAdjustableModelDefinition(
     std::unique_ptr<RiskFactorSexAgeTable> expected,
-    std::unique_ptr<std::unordered_map<core::Identifier, double>> expected_trend)
-    : expected_{std::move(expected)}, expected_trend_{std::move(expected_trend)} {
+    std::unique_ptr<std::unordered_map<core::Identifier, double>> expected_trend,
+    std::unique_ptr<std::unordered_map<core::Identifier, int>> trend_steps)
```
alexdewar, line 204: Am I right in thinking that `expected_trend` and `trend_steps` will have the same keys? If so, would it make sense to just have the one map, with each value being a struct containing both of these values?

src/HealthGPS/risk_factor_adjustable_model.h
```diff
@@ -55,6 +57,11 @@ class RiskFactorAdjustableModel : public RiskFactorModel {
     void adjust_risk_factors(RuntimeContext &context, const std::vector<core::Identifier> &factors,
                              OptionalRanges ranges, bool apply_trend) const;
 
+    /// @brief Gets the number of time steps to apply the trend
+    /// @param factor The risk factor to get the trend steps
+    /// @returns The number of time steps to apply the trend
+    int get_trend_steps(const core::Identifier &factor) const noexcept;
```
alexdewar, line 63: This shouldn't be marked `noexcept`, because the `at()` method will throw a `std::out_of_range` if `factor` doesn't exist.

alexdewar (changes requested):
Mostly looks good to me, but I've got a few questions.

As I've merged #532 now, you'll need to merge in `main` and update the schemas to include the `TrendSteps` properties (sorry!).

src/HealthGPS/risk_factor_adjustable_model.cpp
jamesturner246: Sure. `t` is the smaller of the number of elapsed years and the upper bound on elapsed years, used to bound the exponential growth. Maybe `num_trend_steps` is a better name.

So at the start, when `elapsed <= trend_steps`, we have `expect_new = expect * trend^elapsed`, but after we have `expect * trend^trend_steps`.

src/HealthGPS/risk_factor_adjustable_model.cpp
jamesturner246: Mind If I do this one later? Still not completely certain that a new bunch of vars won't be requested soon.

jamesturner246:
I've opened #544 regarding the schema update. Sorry, should have done it here.

alexdewar (approved):
Given that we're dealing with the schema thing in #544, I think this one is good to merge.

Issue #531, opened 2024-09-25
[BUG] EBHLM model fails to initialise


alexdewar:
## Describe the bug

If you try to use a model with the EBHLM type it fails to initialise with the error: "Risk factor expected trend mapping is empty"

## To Reproduce

- Load any model using the EBHLM dynamic model type (e.g. the `HLM_France`) example

## Additional context

The problem is that the expected trend mapping is initialised to an empty map before being passed into the `DynamicHierarchicalLinearModelDefinition` constructor, which throws an error whenever the map is empty (i.e. every time).

I'm not sure if this is placeholder code with the goal that there will be trend mappings for this model type in future (with extra JSON config options) or if it was just an oversight?

PR #532, opened 2024-09-25
Add new schema for static linear model and enforce model schemas

### LINKED ISSUE #529: Update model schemas to correspond to latest code
Currently the model schemas are out of date with respect to the code, as the time trends work has involved adding some new properties to the files (see #521). We should make some new (v2) schemas for all the bits that have changed.

### LINKED ISSUE #530: Enforce model schemas when loading models
We currently enforce the schema for the main config file, but the dynamic + static config files are not checked against their respective schemas, which we should remedy.

Depends on #529.

alexdewar:
This PR adds a new v2 schema for the static linear model, which adds the various trend-related fields. If you could check that the constraints I've chosen seem sensible (e.g. does anything need a max/min value property?) that would be great. Besides that, I don't *think* any of the other schemas need new versions yet (though see #531), but if I've missed any, please let me know!

I've added the code to enforce these schemas at model load time. I did a small amount of refactoring to facilitate this (mostly amalgamating the static and dynamic model loading functions).

Closes #529. Closes #530.

### DIFF
--- a/schemas/v2/config/models/static.json
+++ b/schemas/v2/config/models/static.json
@@ -1,0 +1,15 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "https://raw.githubusercontent.com/imperialCHEPI/healthgps/main/schemas/v2/config/models/static.json",
+    "anyOf": [
+        {
+            "$ref": "../../../v1/config/models/dummy.json"
+        },
+        {
+            "$ref": "../../../v1/config/models/hlm.json"
+        },
+        {
+            "$ref": "staticlinear.json"
+        }
+    ]
+}
--- a/src/HealthGPS.Input/model_parser.cpp
+++ b/src/HealthGPS.Input/model_parser.cpp
@@ -2,16 +2,18 @@
 #include "configuration_parsing_helpers.h"
 #include "csvparser.h"
 #include "jsonparser.h"
+#include "schema.h"
 
 #include "HealthGPS.Core/exception.h"
 #include "HealthGPS.Core/scoped_timer.h"
 
 #include <Eigen/Cholesky>
 #include <Eigen/Dense>
-#include <algorithm>
-#include <filesystem>
 #include <fmt/color.h>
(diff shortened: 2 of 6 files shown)


src/HealthGPS.Input/model_parser.cpp
```diff
@@ -22,6 +24,56 @@
 #define MEASURE_FUNCTION()
 #endif
 
+namespace {
+// The latest schema version for each model
+constexpr int DummyModelSchemaVersion = 1;
+constexpr int EBHLMModelSchemaVersion = 1;
+constexpr int KevinHallModelSchemaVersion = 1;
+constexpr int HLMModelSchemaVersion = 1;
+constexpr int StaticLinearModelSchemaVersion = 2;
```
jamesturner246, line 33: No need to worry about it for now, but will it be possible eventually to load older configs using older schemas? For instance by using the config's schema version directly?

schemas/v2/config/models/static.json
jamesturner246, line 1: To be clear, this isn't used anywhere right now, right?

jamesturner246 (approved):
No, I think the schema seems sensible. Depending on if you want to, strictly speaking a number of the numbers cannot be negative. For instance stddevs, prevalence,  but this covers the majority of it. Up to you.

Else LGTM :+1: 
Usual questions

schemas/v2/config/models/static.json
alexdewar: I should have said, but no, it's not used by HGPS. I do use `static.json` and `dynamic.json` for validating config files in `healthgps-examples`, but in HGPS I load the model schemas directly because you get clearer error messages that way.

src/HealthGPS.Input/model_parser.cpp
alexdewar: Yep, that's the plan. Atm I'm not requiring the `$schema` attribute, but we could do. I was just thinking that we could try to validate a file with the newest version of the schema and if that doesn't work, we try the next oldest etc.

PR #534, opened 2024-09-25
Fix bundling of DLLs on Windows

### LINKED ISSUE #533: [BUG] Missing DLLs for Windows binaries
As I mentioned in the meeting, I've been having some trouble getting the latest version of HealthGPS to build and run on Windows. While I'm not sure how much of the building stuff is a "me" problem, one of the problems I was facing was that once the binary *was* built, it would fail to run with a message complaining that the TBB DLL couldn't be found. This makes sense: the only DLL in the output folder was `fmt.dll`.

Today I tried the prebuilt binaries which we generate as a CI artifact (specifically [this one](https://github.com/imperialCHEPI/healthgps/actions/runs/10989817972) on the `main` branch). I don't have access to a Windows machine at this moment, but tried it with `wine` and got an error message about a whole spate of missing DLLs:

```
011c:err:module:import_dll Library tbb12.dll (which is needed by L"Z:\\home\\alex\\Downloads\\windows-release\\bin\\HealthGPS.Console.exe") not found
011c:err:module:import_dll Library libzippp.dll (which is needed by L"Z:\\home\\alex\\Downloads\\windows-release\\bin\\HealthGPS.Console.exe") not found
011c:err:module:import_dll Library curlpp.dll (which is needed by L"Z:\\home\\alex\\Downloads\\windows-release\\bin\\HealthGPS.Console.exe") not found
011c:err:module:import_dll Library libcurl.dll (which is needed by L"Z:\\home\\alex\\Downloads\\windows-release\\bin\\HealthGPS.Console.exe") not found
011c:err:module:import_dll Library libcrypto-3-x64.dll (which is needed by L"Z:\\home\\alex\\Downloads\\windows-release\\bin\\HealthGPS.Console.exe") not found
011c:err:module:loader_init Importing dlls for L"Z:\\home\\alex\\Downloads\\windows-release\\bin\\HealthGPS.Console.exe" failed, status c0000135
```

Again, these files aren't bundled with the executable as they should be, so it's no wonder it isn't working.

It seems that the reason we *do* get `fmt.dll` bundled is because of this line: https://github.com/imperialCHEPI/healthgps/blob/main/src/HealthGPS.Console/CMakeLists.txt#L54

I'm not sure whether or not that is the "right" way to do this with CMake. If so, should we be putting all these statements in `src/HealthGPS.Console/CMakeLists.txt` or sprinkling them around in the `CMakeLists.txt` for the different submodules? Dunno.

alexdewar:
Solved using advice from here: https://stackoverflow.com/questions/10671916/how-to-copy-dll-files-into-the-same-folder-as-the-executable-using-cmake/69736197#69736197

Both local builds and the CI generated binaries now work.

I've also added steps to the CI to check that `HealthGPS.Console` will run with the `--help` flag, which should catch these sorts of problem in future.

I'm now hitting another error when actually running the program 😞 , but I'll open a separate issue for this.

Fixes #533.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,8 +97,11 @@ jobs:
       - uses: ammaraskar/gcc-problem-matcher@master
       - name: Build HealthGPS (debug)
         if: '!cancelled()' # Run this step, even if the previous one fails
-        run: |
-          cmake --build --preset=debug-build-${{ matrix.platform }} --target=install
+        run: cmake --build --preset=debug-build-${{ matrix.platform }} --target=install
+
+      - name: Execute HealthGPS (debug)
+        if: '!cancelled()' # Run this step, even if the previous one fails
+        run: out/install/${{ matrix.platform }}-debug/bin/HealthGPS.Console --help
 
       - name: Build HealthGPS (release)
         if: '!cancelled()' # Run this step, even if the previous one fails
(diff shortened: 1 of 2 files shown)


alexdewar:
Actually, I'll make a single PR for all the Windows-related fixes. Don't bother reviewing this.

PR #537, opened 2024-09-27
Various Windows fixes

### LINKED ISSUE #533: [BUG] Missing DLLs for Windows binaries

### LINKED ISSUE #536: [BUG] Zip files are corrupted on download on Windows
## Describe the bug

When downloading a static data file on Windows, the file is always corrupted (the checksum doesn't match). On examining the downloaded file, I found that CRs had been inserted before every LF 🙃.

I'm digging into it now.

## To Reproduce

Run HGPS using any of the examples on Windows.

## Software versions

- OS: Windows 10

alexdewar:
This PR fixes the Windows (MSVC) build of HGPS.

There were a few issues I had in trying to get HGPS to run on Windows:

1. A bunch of dependent DLLs were missing from the install directory, meaning both local and CI builds wouldn't run. Fixed by scanning the exe for depending DLLs and including them.
2. There were a few places where `fstream`s were being opened in text mode rather than binary mode, which won't work with binary files on Windows
3. A trailing slash on the end of directory names was breaking things

Fixes #533. Fixes #536.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,8 +97,11 @@ jobs:
       - uses: ammaraskar/gcc-problem-matcher@master
       - name: Build HealthGPS (debug)
         if: '!cancelled()' # Run this step, even if the previous one fails
-        run: |
-          cmake --build --preset=debug-build-${{ matrix.platform }} --target=install
+        run: cmake --build --preset=debug-build-${{ matrix.platform }} --target=install
+
+      - name: Execute HealthGPS (debug)
+        if: '!cancelled()' # Run this step, even if the previous one fails
+        run: out/install/${{ matrix.platform }}-debug/bin/HealthGPS.Console --help
 
       - name: Build HealthGPS (release)
         if: '!cancelled()' # Run this step, even if the previous one fails
@@ -106,6 +109,10 @@ jobs:
           # Build documentation so we can show Doxygen warnings
           cmake --preset=${{ matrix.platform }}-release -DWARNINGS_AS_ERRORS=ON -DBUILD_DOC=ON
           cmake --build --preset=release-build-${{ matrix.platform }} --target=install
+
+      - name: Execute HealthGPS (release)
+        if: '!cancelled()' # Run this step, even if the previous one fails
+        run: out/install/${{ matrix.platform }}-release/bin/HealthGPS.Console --help
 
       - name: Upload artifacts
         if: matrix.is_main
--- a/src/HealthGPS.Console/CMakeLists.txt
+++ b/src/HealthGPS.Console/CMakeLists.txt
@@ -49,9 +49,28 @@ target_link_libraries(
             cxxopts::cxxopts
             TBB::tbb)
 
-install(TARGETS HealthGPS.Console DESTINATION "")
+install(
+    TARGETS HealthGPS.Console
+    RUNTIME
+    ARCHIVE
+    LIBRARY
+    RUNTIME
+    FRAMEWORK
+    BUNDLE
+    PUBLIC_HEADER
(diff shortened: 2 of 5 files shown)


alexdewar:
Btw: Windows binaries should work fine with `wine`. I had to install the Visual C++ redist library too: `winetricks vcrun2022`

It's handy if you want to test that the CI-generated Windows binaries work without having to boot Windows.

jamesturner246 (approved):
Okay. Not sure I understand the CMake changes, `RUNTIME, ARCHIVE, LIBRARY, RUNTIME, ...`. And esp. the windows part.

Also is the create directory exception not needed anymore?

I trust it's right, so LGTM.

alexdewar:
> Okay. Not sure I understand the CMake changes, `RUNTIME, ARCHIVE, LIBRARY, RUNTIME, ...`. And esp. the windows part.

I don't entirely understand why those bits were needed either :upside_down_face:. I copied from here: https://stackoverflow.com/questions/10671916/how-to-copy-dll-files-into-the-same-folder-as-the-executable-using-cmake

The bit about installing `RUNTIME_DEPENDENCIES` just figures out what DLLs are needed for the EXE, excluding system libraries, and copies them to the install folder.

> Also is the create directory exception not needed anymore?

I was getting spurious failures here, I think because the extra `.` causes Windows to try to create a directory twice which fails. Yuck. Anyway, it works fine if we just assume that directories are created successfully and raise errors if creating files fails.

PR #544, opened 2024-10-03
Add new params to JSON schemas.

### LINKED ISSUE #543: [BUG] schemas for KH and staticlinear params need syncing
## Describe the bug

New parameters have been added to both staticlinear and kevinhall models, and the new JSON schemas must be updated accordingly.

## To Reproduce

Run model and see failed schema validation.

## Expected behavior

Both schemas must be modified to check for these new JSON params.

## Software versions

N/A

## Additional context

`healthgps/schemas/v1/config/models/kevinhall.json`
`healthgps/schemas/v2/config/models/staticlinear.json`

jamesturner246:
Resolves #543 

Adds missing new params to KH and staticlinear JSON schemas.

### DIFF
--- a/schemas/v1/config/models/kevinhall.json
+++ b/schemas/v1/config/models/kevinhall.json
@@ -63,12 +63,20 @@
                                 "type": "null"
                             }
                         ]
+                    },
+                    "ExpectedTrend": {
+                        "type": "number"
+                    },
+                    "TrendSteps": {
+                        "type": "integer"
                     }
                 },
                 "required": [
                     "Name",
                     "Nutrients",
-                    "Price"
+                    "Price",
+                    "ExpectedTrend",
+                    "TrendSteps"
                 ],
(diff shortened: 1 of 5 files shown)


alexdewar (changes requested):
The changes that you've made all look correct, **BUT** we want a v2 of the `kevinhall.json` schema, not a modified v1. If we just change v1, then all the existing config files out there (including the ones in the examples repo) will start failing validation, even though they're well formed files (for old versions of HGPS, anyway).

You need to do what I did with the `staticlinear.json` schema and copy it over to the v2 folder, then update any `$ref` properties to point to the right subschemas (in the v1 folder) before making changes.

You'll also need a v2 of `dynamic.json` (see what I did for `static.json` for an example).

jamesturner246:
Ah I see. Done :+1:

alexdewar (approved):
LGTM :+1:

PR #546, opened 2024-10-04
Handle missing trend vars in EBHLM.

### LINKED ISSUE #531: [BUG] EBHLM model fails to initialise

jamesturner246:
Resolves #531 

Introduces placeholder trend parameters into the EBHLM model, until trending is implemented there.

### DIFF
--- a/src/HealthGPS.Input/model_parser.cpp
+++ b/src/HealthGPS.Input/model_parser.cpp
@@ -398,6 +398,11 @@ load_ebhlm_risk_model_definition(const nlohmann::json &opt, const Configuration 
 
     auto info = LiteHierarchicalModelInfo{};
 
+    // Risk factor expected values by sex and age.
+    auto expected = load_risk_factor_expected(config);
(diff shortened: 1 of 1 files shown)


alexdewar (approved):
If it works, I'm ok with it. But couldn't you just make these parameters optional for now instead? You can make them non-optional once you've implemented it properly for this model. (Obvs this isn't worth doing if it's too much of a faff.)

jamesturner246:
Think I'll leave this one. May well be refactoring it anyway, so will figure it out then.

PR #548, opened 2024-10-10
Macos2

alexdewar:
I've rebased these commits to see if vcpkg etc. has magically fixed itself since I last tried.

Closes #155.


alexdewar:
Still broken. Let's shelve this for now

PR #550, opened 2024-10-11
Add `--dry-run` flag to validate input files without running simulation

### LINKED ISSUE #549: [ENHANCEMENT] Allow for validating input files without starting actual simulation

For HPC users (i.e. @jzhu20), it would be helpful to know that input files can be parsed correctly by HealthGPS before actually submitting a job. 

Add a flag to check the input files load correctly without actually starting the simulation. How about `--dry-run`?

## Describe alternatives you've considered

While you *can* run HealthGPS then, if the files validate correctly, just kill the program when the simulation starts, this is a bit ugly.

alexdewar:
This PR adds a new `--dry-run` flag which can be passed to HGPS in order to validate input files without actually running a simulation. The idea is that it will enable @jzhu20 to check that the input data files are valid before submitting HPC jobs.

Closes #549.


jamesturner246 (approved):
LGTM. Great! This will help solve some headaches :+1:

PR #553, opened 2024-10-30
Tidy up and fix docs

alexdewar:
I've had a go at tidying up the docs, in particular the installation instructions and the out-of-date instructions related to input files. I've also done some generic tidying (headers, reformatting tables etc.) over all of the docs. (Sorry for mixing the two -- they probably should have been separated out.)

If you just want to see the content changes, `getstarted.md` and `userguide.md` are where I actually added/removed sections.

Other changes:

- Swap `markdownlint` hook for `markdowlint-fix` and put config in separate file
- Signpost @SaranjeetKaur's R package
- Fix markdown headings (heading level -- and some were numbered for some reason)
- Remove duplicated content in user guide about building HGPS
- Fix some broken links (we don't have a link checker running as part of our CI seemingly)
- Reformat tables with [markdown-table-formatter](http://markdowntable.com/)

Things not addressed:

- I'm not sure if the example R code in the docs still works. I've signposted the R package, but I don't know whether the documentation should just tell users to use that instead?
- I'm not sure of the value of a lot of the HPC content. If we want to keep it we should probably go through it and check it's not out of date. I personally couldn't get the EasyBuild version of HealthGPS working and in any case that would represent a very old version now anyway, so maybe we don't want to tell users to do this? Anything specific to Imperial's HPC doesn't really belong in the HGPS repo -- we should refer users to the RCS wiki (or whatever) instead.

Any thoughts welcome!


docs/getstarted.md
```diff
@@ -96,12 +89,10 @@ p <- ggplot(data=df, aes(x=time, y=bmi, group=interaction(scenario, gender))) +
...
-|![Experiment BMI Projection](images/bmi_projection.svg)|
-|:--:|
-|*Experiment BMI projection example*|
+| ![Experiment BMI Projection](images/bmi_projection.svg) |
+|:-------------------------------------------------------:|
+|           *Experiment BMI projection example*           |
 
 In a similar manner, the resulting dataset `df`, can be re-created and expanded to summarise other variables of interest, create results tables and plots to better understand the experiment.
 
 [configjson]:https://github.com/imperialCHEPI/healthgps/blob/main/example/France.Config.json "Configuration file example"
```
jamesturner246, line 98: Dead link!

docs/getstarted.md
```diff
@@ -4,40 +4,33 @@
...
+## Results
 
-# Results
+NB: For analysis and visualisation of *Health GPS* output data, we recommend the [`healthgpsrvis` R package](https://imperialchepi.github.io/healthgpsrvis/), but here we show how to do some analysis manually.
 
 The current model output format is JSON (JavaScript Object Notation), an open standard file format designed for data interchange in human-readable text. It is language-independent; however, all programming language and major data science tools supports JSON format because they have libraries and functions to read/write JSON structures. To read the model results in [R](https://www.r-project.org/), for example, you need the [`jsonlite`](https://cran.r-project.org/web/packages/jsonlite/vignettes/json-aaquickstart.html) package:
```
jamesturner246, line 35: I'd probably emphasize that the main output is CSV. We haven't ever used the JSON output. and I'm not convinced it is correct. 

Relatedly, the R analysis example is probably inaccurate, and thus misleading..

jamesturner246 (changes requested):
Looks like it covers all the input layer stuff.

Just one thing about the output data notes -- we haven't actually used the JSON output in the past, and I'm not convinced it is correct.

docs/getstarted.md
alexdewar: Good spot!

docs/getstarted.md
alexdewar: > I'd probably emphasize that the main output is CSV. We haven't ever used the JSON output. and I'm not convinced it is correct.

Sure, will do.

> Relatedly, the R analysis example is probably inaccurate, and thus misleading..

That was my worry... Shall I make an issue for it? Perhaps @SaranjeetKaur could take a look at some point, if she has time. If it *is* inaccurate, we can just delete all that content and signpost the R package instead.

jamesturner246: Yeah, sounds good. Her stuff is less likely to be wrong, and a shame not to use it.

jamesturner246 (approved):
LGTM! :+1:

PR #580, opened 2025-03-10
Updated the modelling.json schema to handle the demographic models for region and ethnicity

M-Ghosh9:
To include the new demographic models for region and ethnicity where region is assigned based on age/gender probabilities and ethnicity is assigned based on age/gender/region probabilities. 

Hence I've updated the schema to take these new coefficients.

### DIFF
--- a/schemas/v1/config/modelling.json
+++ b/schemas/v1/config/modelling.json
@@ -34,12 +34,18 @@
                                 "coefficients": {
                                     "type": "object",
                                     "properties": {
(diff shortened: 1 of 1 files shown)


M-Ghosh9:
The v1/config has rules for the config.json made by the user. 
Inside that we have inputs, modelling, running and outputs. 
I've made change to modelling to handle the new demographic models for region and ethnicity

dlaydon (approved):
Approved after discussion

PR #581, opened 2025-03-11
Fix schema validation for demographic_models property

M-Ghosh9:
Added demographic_models to the required properties in the modelling schema to fix validation errors when using this property.

The current required contains- 
"required": [
    "ses_model",
    "demographic_models",  // added this
    "risk_factors",
    "risk_factor_models",
    "baseline_adjustments"
],

This is coz despite adding a demographic_models to the v1/config/modelling.json, it is missing in the required section hence I'm getting this error of -
_Invalid configuration - /modelling/demographic_models: Additional property 'demographic_models' not allowed by schema._.

This should solve the above error.

### DIFF
--- a/schemas/v1/config/modelling.json
+++ b/schemas/v1/config/modelling.json
@@ -258,6 +258,7 @@
     },
     "required": [
         "ses_model",
+        "demographic_models",
         "risk_factors",
         "risk_factor_models",
         "baseline_adjustments"


M-Ghosh9:
This should fix the demographic_models not allowed by updating the modelling schema.

dlaydon (approved):

M-Ghosh9:
Why did Mahima fail before? 

The application, after building was using the schemas available in the local build directory. As a result, when I was making changes and updating main, the local build directory's schemas were still the same. I needed to update those too as my build was not detecting the schemas. Probably some error with my CMake so I'll have to do this neat trick. 

But lesson learnt- Update everywhere. There is no such thing as being too cautious.

M-Ghosh9 (commented):
Schema now accepts the new demographic_models that contain the region and ethnicity probabilities and coefficients.

PR #589, opened 2025-04-14
Clean FINCH code for region, ethnicity, income and physical activity

M-Ghosh9:
This is a cleaner version of the feature/finch-debug where I've removed all the excess try-catch that were masking the actual problems- given the deadline.


M-Ghosh9:
Does data loading for the risk factor using the .csv files instead of JSON. This reduces the copying and pasting redundancy.

PR #590, opened 2025-04-14
Dj lparallel 2

dlaydon:
Hi @jamesturner246 and @alexdewar 

I've made a few changes to use Thread Building Blocks more often throughout the code. I've done this only in places where it's thread-safe, and not anywhere where vectors are "grown" (e.g. push_back or emplace). Works fine for me locally, but I'd appreciate your thoughts and checking. 

Thanks!


M-Ghosh9:
Outdated

PR #605, opened 2025-07-24
India/income trend

M-Ghosh9:
1. Added new variable income trend
2. Added initialise_income_trend in static_linear_model
3. Added update income_trend in static_linear_model
4. Updated model_parser.cpp to accept trend types- null, trend or income_trend


jzhu20 (commented):
It looks good.

PR #606, opened 2025-07-29
Finch/fat

M-Ghosh9:
Traces fat calculation the 1st year of the simulation for a 30 year old woman.

// Tracing configuration - easily modifiable
static constexpr bool ENABLE_FAT_TRACING = true;  // Toggle tracing on/off
static constexpr int TARGET_AGE = 30;             // Target age
static constexpr core::Gender TARGET_GENDER = core::Gender::female;  // Target gender
static const core::Identifier TARGET_RISK_FACTOR = "FoodFat"_id;        // Target risk factor

You can change which risk factor you want to track for what age and gender automatically. Right now it does for risk factor FoodFat.

### DIFF
--- a/src/CMakeSettings.json
+++ b/src/CMakeSettings.json
@@ -1,0 +1,15 @@
+﻿{
+  "configurations": [
+    {
+      "name": "x64-Debug",
+      "generator": "Ninja",
+      "configurationType": "Debug",
+      "inheritEnvironments": [ "msvc_x64_x64" ],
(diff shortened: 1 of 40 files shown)


M-Ghosh9:
Now it contains-
1. Franco's new change to calculate factor without multiplying with std dev
2. New rows for Cholesky before and after

P.S. Has merge conflict. But no worries- it's just for testing

PR #620, opened 2025-09-04
Finch/correlation

M-Ghosh9:
Ordering the names_ vector to exact order as the correlation matrix order


M-Ghosh9:
Outdated. Already updated and merged in Main

PR #622, opened 2025-09-11
Finch/RF Adjust

M-Ghosh9:
Contains the new adjustments to risk factors after correcting the names_ to have the exact order of the correlation matrix.

1. added a vector other_risk_factor_names that handles other risk factors such as height, weight, physical activity, income
2. calling separate for adjust to risk factors (ones for names_ which contain food and second for other_risk_factor_names that contain height, weight, income, etc)
3. turns out the invalid mapping error shows that with each call of the adjust_risk_factor function, the map was being recalculated hence there was  a mismatch between the previous 21 risk factors that are food related and the 2nd 5 other risk factors. 
4. so I combined all the risk factors (21+ 5) before calling adjust_risk_factor. 
5. code proceeded from there but then got stuck between baseline and intervention.  

The code runs fully with correct adjustments to factors mean for income and physical activity

### DIFF
--- a/src/CMakeSettings.json
+++ b/src/CMakeSettings.json
@@ -1,0 +1,15 @@
+﻿{
+  "configurations": [
+    {
+      "name": "x64-Debug",
+      "generator": "Ninja",
+      "configurationType": "Debug",
+      "inheritEnvironments": [ "msvc_x64_x64" ],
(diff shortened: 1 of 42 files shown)


M-Ghosh9:
Already updated in Main

PR #648, opened 2026-04-20
added storage for income based factors mean files

M-Ghosh9:
Phrase 2 of the https://github.com/imperialCHEPI/healthgps/blob/main/income_quintile_factor_means-plan.md. 

What I changed-

1. In src/HealthGPS.Input/model_parser.cpp:

- Added a shared loader helper for expected tables:
- loads male/female CSV pair,
- validates required age coverage,
- throws clear source-tagged errors (overall, or income stratum '<id>').
- Kept existing overall load path, now routed through that helper.
- read baseline_adjustment.income_stratum_factors_mean,
- load each stratum table pair when enabled,
- validate that every stratum table contains all required risk factors (male+female),
- fail fast with stratum-specific context if anything is missing.
- Passes loaded stratum tables + enabled flag + adjustment stratum count into StaticLinearModelDefinition.

2. In src/HealthGPS/static_linear_model.h:

- Added IncomeStratumExpectedTableEntry alias.
- Extended constructors for StaticLinearModel and StaticLinearModelDefinition to accept:
- ordered per-stratum expected tables,
- enabled flag,
- adjustment stratum count.

3. In src/HealthGPS/static_linear_model.cpp:

- Updated constructor definitions and create_model() forwarding to carry these new stratum inputs through.
- Added MAHIMA comments clarifying this is Phase 2 Step 1 data plumbing.


M-Ghosh9:
What I changed
1) Added stratum-aware adjustment plumbing in RiskFactorAdjustableModel
In src/HealthGPS/risk_factor_adjustable_model.h and src/HealthGPS/risk_factor_adjustable_model.cpp:

- Extended adjust_risk_factors(...) with optional parameters:
- expected_override (use a specific expected table instead of overall),
- income_stratum_filter (apply only to people in stratum k).
- Threaded the same options through:
- calculate_adjustments(...),
- calculate_simulated_mean(...).
- Apply phase now skips non-matching people when stratum filter is set.
- Simulated mean also only uses the filtered stratum population.
- Kept defaults so all existing non-stratum call sites still behave exactly as before.

2) Implemented Phase 2 runtime order in StaticLinearModel (generate + update)
In src/HealthGPS/static_linear_model.cpp:

For initial (non-trended) adjustment when stratum mode is enabled:

- Overall income-only adjustment (if project requirements enable income adjustment),
- Assign N income adjustment strata by rank,
- Per-stratum RF adjustment using that stratum’s expected table,
- Per-stratum PA adjustment using that stratum’s expected table,
- Final income categories still assigned later via existing income_categories_ flow.
- For trended adjustment when stratum mode is enabled:
- 
- Overall income trended pass (if enabled),
- Recompute strata from updated income,
- Per-stratum RF trended adjustment,
- Per-stratum PA trended adjustment,
- Existing final category rebucketing remains in place.
- When stratum mode is off, the legacy combined path remains unchanged.

3) Efficiency choices

- Stratum filtering happens inside the adjustment engine; no population copies.
- Existing parallel_for_each apply path is preserved.
- Factor/range vectors are built once per stage, reused across strata loops.
- Only configured/needed factors run (income-only, RF-only, PA-only split).

MariiaMurasheva (approved):

M-Ghosh9:
What I changed
1) Richer sampled debug rows (not just one factor + age 0)
I updated the row capture path in risk_factor_adjustable_model so each stratum-adjustment call now collects a bounded pseudo-random sample across:

different risk factors,
both sexes,
multiple ages (when present),
with expected/simulated/delta/current/after-delta/final values.
Files:

src/HealthGPS/risk_factor_adjustable_model.h
src/HealthGPS/risk_factor_adjustable_model.cpp


2) Clear “after delta” vs “final clamped”
The debug row now tracks both:

Current: value before applying delta
AfterDelta: Current + Delta (before clamp)
Final: value after clamp/actual stored final value
So you can now see exactly what clamping changed.

3) Include which income-stratum file was used
I added StratumID to each debug row and populate it from income_stratum_expected_tables_[k].first in all four paths:

initial
trended
yearly-initial
yearly-trended
That tells you which stratum table pair was used for that row (your “which of the 10 files” request via table ID mapping).

src/HealthGPS/static_linear_model.cpp


4) Updated output table columns
[MAHIMA][INCOME-STRATUM DELTA/APPLY EXAMPLES] table now prints:

Bucket
StratumID
PersonID
Factor
Sex
Age
Expected
SimulatedMean
Delta
Current
AfterDelta
Final
So it stays the same style but with richer diagnostics.

PR #654, opened 2026-05-07
configuration updates for windows and linux

M-Ghosh9:
After reboot:

cl.exe is not on PATH
Ninja does not auto-load MSVC
So CMake fails

-------> Visual Studio + Ninja already knows:

Which MSVC to use
Which host/target architecture to use
How to inject the environment correctly

"CMAKE_C_COMPILER": "cl",
"CMAKE_CXX_COMPILER": "cl",
"CMAKE_LINKER": "link",
By removing those 3 lines from CMakePresets.json, you let VS do its job every time, even after reboot.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -54,9 +54,6 @@
             },
             "cacheVariables": {
                 "CMAKE_BUILD_TYPE": "Debug",
-                "CMAKE_C_COMPILER": "cl",
-                "CMAKE_CXX_COMPILER": "cl",
-                "CMAKE_LINKER": "link",
                 "VCPKG_TARGET_TRIPLET": "x64-windows",
                 "CMAKE_TOOLCHAIN_FILE": {
                     "value": "$env{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake",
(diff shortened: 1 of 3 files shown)


M-Ghosh9:
Fix 1-
Removed CMAKE_C_COMPILER, CMAKE_CXX_COMPILER, and CMAKE_LINKER
Let Visual Studio + vcpkg choose MSVC automatically
Deleted the build directory and reconfigured

Result: CMake configure succeeded again, reliably across restarts

M-Ghosh9:
Fix 2-

CMake config is robust (no reliance on cl being on PATH)
MSVC uses proper C++ exception semantics everywhere (/EHsc)
Clean Windows Release builds
Safe for GitHub CI (Windows + Linux)
No weakening of warnings
No edits to vcpkg or third‑party code

M-Ghosh9:
2 classic CMake/MSVC pitfalls:

Manually forcing compilers with Ninja (fragile across restarts)
Compiler flags applied too late or to the wrong targets

I fixed both the right way, using:
- Presets instead of environment assumptions
- Target‑scoped compile options instead of global hacks

jzhu20 (approved):

PR #660, opened 2026-06-02
unique ID

M-Ghosh9:
Note: IDs are unique within one run. Baseline and intervention each have their own Population and counter; initial IDs 1..N still mean “same starting person,” not one global ID across both runs. Tracking death in exports still needs year + active status (or last-seen year), because a dead person keeps their ID in the slot until it is recycled.

### Before (implemented slot-based ID reuse)

```mermaid
flowchart TD
  startA[PersonInSlot_i_ID_iPlus1] --> deathA[PersonDiesOrEmigrates]
  deathA --> recycleA[Slot_iMarkedRecyclable]
  recycleA --> newbornA[NewPersonPlacedInSlot_i]
  newbornA --> sameIdA[AssignedID_iPlus1_Again]
  sameIdA --> mixedA[SameIDMapsToDifferentPeopleOverTime]
```

### Updated (lifetime-unique IDs with slot reuse only)

```mermaid
flowchart TD
  initB[InitialSlot_i_GetsID_iPlus1] --> counterB[next_person_id_InitialisedTo_NPlus1]
  counterB --> deathB[PersonDiesOrEmigrates]
  deathB --> recycleB[Slot_iReusedForMemoryOnly]
  recycleB --> newPersonB[NewPersonPlacedInRecycledOrNewSlot]
  newPersonB --> assignB[AssignID_next_person_id_ThenIncrement]
  assignB --> uniqueB[IDNeverReused_LifetimeUnique]
```

### DIFF
--- a/Technical Documentations/same_person_id_across_baseline_and_intervention-plan.md
+++ b/Technical Documentations/same_person_id_across_baseline_and_intervention-plan.md
@@ -1,7 +1,21 @@
 ---
 name: Same person ID across baseline and intervention
 overview: Assign Person IDs from population index (id = index + 1) so the same logical person has the same ID in both baseline and intervention runs. Changes are limited to Person and Population with MAHIMA comments; no change to scenario logic or event bus.
-todos: []
+todos:
+  - id: confirm-baseline-sync-contract
+    content: Confirm and document baseline -> intervention sync contract (aggregate tables only, one-way) in the plan and implementation notes.
+  - id: add-population-next-id-counter
+    content: Add Population monotonic next_person_id_ state initialised to initial_size + 1 after initial population construction.
+  - id: switch-post-initial-id-assignment
+    content: Update Population add and add_newborn_babies so all post-initial entrants get ID from next_person_id_++ on both recycled-slot and append paths.
+  - id: preserve-initial-id-alignment
+    content: Keep initial ID assignment as i + 1 to preserve baseline/intervention initial cohort comparability.
+  - id: refresh-tests-lifetime-unique
+    content: Replace slot-reuse ID assertions with lifetime-unique assertions in Population tests, including recycled-slot replacement cases.
+  - id: validate-individual-tracking-behavior
+    content: Validate tracking output expectations (no demographic identity swapping under one ID due to slot reuse) with targeted run/test checks.
+  - id: performance-memory-check
+    content: Record runtime and memory impact checks for large runs (including 14M-population assumptions and observed deltas).
 isProject: false
 ---
 
@@ -16,6 +30,13 @@ Make the same logical person have the **same ID** in both baseline and intervent
 - **Scenario code** (physical_activity, marketing, food_labelling, fiscal): Each scenario holds its own `interventions_book`_ keyed by `entity.id()`. Baseline and intervention are separate Simulation/Scenario instances, so they have separate maps. Same ID in both runs does not cause collision.
 - **No global ID key**: No code keys data globally by Person ID across runs; IDs are only used within one population.
 - **result_file_writer.cpp** `message.id()` is `ResultEventMessage::id()` (event type enum), not Person ID — no change.
+
+## What data is transferred 
+Only aggregate tables, not person-level records:
+
+NetImmigrationMessage = age x gender net migration table
+ResidualMortalityMessage = age x gender residual mortality rates
+No person object, no ID, no region/ethnicity individual records are transferred between scenarios.
 
 ## Design: ID assignment rules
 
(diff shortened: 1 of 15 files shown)


M-Ghosh9:
Initial people (at t = 0): slot i gets ID i + 1 (deterministic, same across baseline/intervention for the starting cohort).

Death / emigration: the person becomes inactive (is_active() false). The slot is not given a new ID yet; the old Person (with the old ID) still sits in that slot until recycling is allowed.

Births / immigration (add, add_newborn_babies): when a dead/emigrated slot is reused, the new person gets next_person_id_++, not the old slot ID. Immigration goes through add() after partial_clone_entity() (clone does not copy the source ID; add assigns a new one).

There is a unit test that encodes this: after death/emigration in slots 3 and 7, newborns in those slots get IDs 11 and 12, not 4 and 8 (Population.Test.cpp, PersonIdInitialDeterministicAndPostInitialLifetimeUnique).

M-Ghosh9:
For one `Population` instance (one scenario run), the intended rule is:

| Question | Intended answer |
| -- | -- |
| Can two active people share the same ID? | No (counter only increases for new entrants; initial IDs 1..N are unique). |
| If someone dies, is their ID given to someone else? | No — the next occupant of that memory slot gets a new ID. |
| Is memory reused? | Yes — same `std::vector` index, new `Person` object. |
| Runtime cost | Essentially unchanged: one extra `size_t` per population + increment per birth/immigrant. |

That matches the intended model: **track by ID while they exist; when they leave, that ID is retired; slot memory is recycled, identity is not.**

M-Ghosh9:
2022: ID 7333 = 45-year-old man (baseline and intervention)  
2023: ID 7333 = 55-year-old woman (baseline) and 45-year-old man (intervention)

| Piece | Plausible? |
| -- | -- |
| 2022: same ID, same 45yo man in both runs | Yes, if 7333 is the same initial or same-order entrant and inputs align |
| 2023 intervention: still 45yo man | Unusual if time advanced one year (should be 46) unless time didn’t step or export is wrong |
| 2023 baseline: 55yo woman vs 2022: 45yo man, same ID, same baseline run | No — not valid for one lifetime under current rules |
| 2023: different age/gender baseline vs intervention at same ID | Possible for high IDs (different entrants); expected for diverging matched starters (e.g. still man in both but different age/death) |

### What “unique throughout their life” means here

| Scope | Guarantee |
| -- | -- |
| One run (baseline or intervention) | Each ID maps to at most one lifetime; no reuse after death/emigration; same ID next year = same person (aging), not a new woman replacing a man |
| Across years (same run) | Track with `person_id + year` (and usually active // death time) |
| Across baseline vs intervention | Only initial IDs 1..N are meant as “paired starters”; higher IDs are not automatically the same person in both runs |

M-Ghosh9:
### What baseline sends to intervention (not “the people”)

Sync is **one-way, aggregate, per year** — not person records.

| Message | Baseline | Intervention |
| -- | -- | -- |
| Net immigration | Computes age × gender net migration table, sends it | Receives table, applies counts to its own population |
| Residual mortality | Computes age × gender residual death rates, sends | Receives rates, uses them in its own death step |

M-Ghosh9:
| Test | What it checks |
| -- | -- |
| InitialCohortIdsAreConsecutiveFromOne | After `Population(N)`, IDs are exactly 1 … N (matches constructor debug checks). |
| MahimaNewEntrantIdsAreNotAlreadyAssigned | After death/emigration + 2 newborns + 1 immigrant, every new ID was not in the set before placement; final IDs are 7, 8, 9 for `init_size = 6`; set size = `init_size + 3`. |
| PersonIdInitialDeterministicAndPostInitialLifetimeUnique | Comment links to the debug guard. |
| AllPopulationIdsArePairwiseDistinct | Unchanged logic; still the duplicate-ID regression. |

jzhu20:
One question please: do people have the same ID at baseline and intervention? @M-Ghosh9

M-Ghosh9:
> One question please: do people have the same ID at baseline and intervention? @M-Ghosh9

Hi @jzhu20 I fixed ID reuse on recycled slots: each person keeps a unique ID for their whole life in that run; only memory slots are reused. 

For the people who are there at the start of the simulation — yes. They get the same ID in baseline and intervention (person 1, person 2, … up to the starting population size). That’s so we can compare the same starting individuals across the two runs.

For anyone born or who immigrates later — no, not necessarily. Baseline and intervention each run their own population (some maybe be alive or dead coz of policies), so the same ID number later on does not always mean the same person in both runs.

Within one run (baseline only, or intervention only), each person keeps one unique ID for their whole life; that ID is not given to someone else after they die.

jzhu20:
> > One question please: do people have the same ID at baseline and intervention? @M-Ghosh9
> 
> Hi @jzhu20 I fixed ID reuse on recycled slots: each person keeps a unique ID for their whole life in that run; only memory slots are reused.
> 
> For the people who are there at the start of the simulation — yes. They get the same ID in baseline and intervention (person 1, person 2, … up to the starting population size). That’s so we can compare the same starting individuals across the two runs.
> 
> For anyone born or who immigrates later — no, not necessarily. Baseline and intervention each run their own population (some maybe be alive or dead coz of policies), so the same ID number later on does not always mean the same person in both runs.
> 
> Within one run (baseline only, or intervention only), each person keeps one unique ID for their whole life; that ID is not given to someone else after they die.

I understood before that even those newborns or immigrants or people who died at baseline but not intervention would keep their ID (i.e. the ID number) throughout the whole simulation/life course. But the empty slot would be recycled. For instance, ID-100 died at baseline, but alive under intervention. This ID-100 will be kept for this person, not used for anyone else at baseline. Am I wrong?

M-Ghosh9:
> > > One question please: do people have the same ID at baseline and intervention? @M-Ghosh9
> > 
> > 
> > Hi @jzhu20 I fixed ID reuse on recycled slots: each person keeps a unique ID for their whole life in that run; only memory slots are reused.
> > For the people who are there at the start of the simulation — yes. They get the same ID in baseline and intervention (person 1, person 2, … up to the starting population size). That’s so we can compare the same starting individuals across the two runs.
> > For anyone born or who immigrates later — no, not necessarily. Baseline and intervention each run their own population (some maybe be alive or dead coz of policies), so the same ID number later on does not always mean the same person in both runs.
> > Within one run (baseline only, or intervention only), each person keeps one unique ID for their whole life; that ID is not given to someone else after they die.
> 
> I understood before that even those newborns or immigrants or people who died at baseline but not intervention would keep their ID (i.e. the ID number) throughout the whole simulation/life course. But the empty slot would be recycled. For instance, ID-100 died at baseline, but alive under intervention. This ID-100 will be kept for this person, not used for anyone else at baseline. Am I wrong?

@jzhu20 You’re right for the starting population (IDs 1 up to the initial size).

Example — ID 100 at start

Baseline: ID 100 is that starter; if they die, ID 100 is not given to anyone else in baseline. The slot may be reused for a newborn, but that newborn gets a new ID (e.g. 100,001), not 100.
Intervention: ID 100 is the same starter for comparison. They can still be alive there even if they died in baseline. Still ID 100 in intervention — not a different person.
So for starters: same ID number in baseline and intervention = same initial person, and within each run that ID is kept for that person’s whole life (not reused after death). 

The fix was mainly about not reusing IDs inside one run when slots are recycled — not about changing how starter IDs line up across baseline and intervention.

jzhu20 (approved):

